PR #53, opened 2015-02-25
feat: enable source maps based on webpack config

mtscout6:
Source-maps do not work for CSS when you base64 encode them, yet it
would be nice to to facilitate dynamically applying source-maps using
`webpack -d`. This will apply the sourceMap query param to the
css-loader if a devtool option is selected in the webpack config.

Relates to webpack/less-loader#36
Relates to webpack/css-loader#40

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+indent_style = tab
+indent_size = 2
+
+# Matches the exact files either package.json or .travis.yml
+[{package.json,.travis.yml}]
+indent_style = space
+indent_size = 2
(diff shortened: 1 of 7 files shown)


mtscout6:
I also added a few unit tests and the necessary .travis.yml and `npm test` script to flip the swicth on for travis-ci.

mtscout6:
@sokra Any thoughts on this? Is there a better way to do this or does this seem like a good track to follow?

sokra:
The idea is good... but I need to find another way to do this... Normally you are not allowed to access the options in a plugin.

mtscout6:
I wonder if it would be better to apply something like this within the [`extract`](https://github.com/webpack/extract-text-webpack-plugin/blob/master/index.js#L42) method? Since that's more along the lines of defining the loaders anyways? The only problem is that it has no context to the active configuration... :confused:

michael-ciniawsky:
@mtscout6 Do you still have interest in this PR, durung to age it targets webpack 1 and would likely need additional changes in conjunction to a rebase, to work with webpack 2. Please let me now how to proceed here :)

michael-ciniawsky:
Closing due to inactivity and age, but feel free to reopen :)

PR #77, opened 2015-06-12
Do not reorder modules in chunks

### LINKED ISSUE #73: Order of CSS is wrong after incremental compilation
CSS order is changed upon incremental recompilation triggered by a change to a CSS file. It no longer matches the order of the initial compilation.

Versions:

* `extract-text-webpack-plugin@0.8.1`
* `webpack@1.9.10`
* `webpack-dev-server@1.9.0`

For example:

```js
// MyComponent.js

import View from '../ui-View';
import './index.css';

// etc
```

Initial compilation produces:

```css
.View { ... }
.MyComponent { ... }
```

Saving a change to `index.css` triggers a rebuild and produces:

```css
.MyComponent { ... }
.View { ... }
```

jbpros:
This potentially fixes #73.

I'm not sure this is the right thing to do as I don't fully understand the big picture. This, however, works for us and fixes the issue where changed files are prepended to the output instead of inserted at the right original location.


sokra:
fixed by https://github.com/webpack/webpack/commit/e20c4859286843fbf37941541693e00f4ae66238

PR #81, opened 2015-07-03
feat: allow to log invalid orders as warnings instead of errors via option

### LINKED ISSUE #80: `chunks` option
We are migrating from a huge legacy codebase. Currently we have a static build for CSS, which basically takes a list of CSS files, concatenates them and minifies them.

We created entries for each of these groups, which look like this:

``` js
// ...
'bla/myCSSblobA': ['a/foo.css','b/xyz.css', 'c/another.css'],
'bla/myCSSblobB': ['a/foo2.css','b/xyz2.css', 'c/another2.css'],
// ...
```

and use the `ExtractTextPlugin` to generate a `bla/myCSSblobA.css` output, etc.
Unfortunately the ExtractTextPlugin extracts CSS from all our entry chunks, but we only need this for our legacy code. For the new code, we want to use the inline CSS. If we provided a `chunks` option (same functionality as for the `CommonsChunkPlugin`) which allowed to control from which chunks to extract and from which chunks not, would you be willing to merge it?

cc @timse @p0wl

p0wl:
When migrating from a large legacy codebase to Webpack, we noticed that many of our CSS group definitions have inconsistent ordering. After trying to fix that ordering for a while, we found out that it was actually not that simple. Especially because the order might be deliberately different per entry-point and there was no way of knowing that without asking the author(s) of the respective CSS group.
The build with the `extract-text-plugin` would still generate the correct (in the sense of the definition given by the developers) output, except the CI would fail because webpack would exit with a non-zero state because of the order "errors".
Therefore we introduced an option to the plugin that allows us to degrade those errors to warnings, which makes the developers aware of their inconsistent ordering without breaking the CI build immediately.

fixes #80

cc @joscha

### DIFF
--- a/index.js
+++ b/index.js
@@ -284,8 +284,9 @@ ExtractTextPlugin.prototype.apply = function(compiler) {
 				if(extractedChunk.modules.length) {
 					extractedChunk.modules.sort(function(a, b) {
 						if(isInvalidOrder(a, b)) {
-							compilation.errors.push(new OrderUndefinedError(a.getOriginalModule()));
-							compilation.errors.push(new OrderUndefinedError(b.getOriginalModule()));
+							var targetLog = options.relaxInvalidOrder ? compilation.warnings : compilation.errors;
+							targetLog.push(new OrderUndefinedError(a.getOriginalModule()));
+							targetLog.push(new OrderUndefinedError(b.getOriginalModule()));
 						}
 						return getOrder(a, b);
 					});


p0wl:
any objections @sokra?

p0wl:
we really would like to see this merged. can you please have a look?

christian-meyer-jimdo:
:+1:

index.js
```diff
@@ -284,8 +284,9 @@ ExtractTextPlugin.prototype.apply = function(compiler) {
 				if(extractedChunk.modules.length) {
 					extractedChunk.modules.sort(function(a, b) {
 						if(isInvalidOrder(a, b)) {
-							compilation.errors.push(new OrderUndefinedError(a.getOriginalModule()));
-							compilation.errors.push(new OrderUndefinedError(b.getOriginalModule()));
+							var targetLog = options.relaxInvalidOrder ? compilation.warnings : compilation.errors;
```
michael-ciniawsky: This could go hand in hand with the `ignoreOrder` option, instead of adding a new option

joscha: Happy to adapt the PR @michael-ciniawsky

michael-ciniawsky: @joscha Yep please go ahead , not :100: my assumption is yet fully correct :)

michael-ciniawsky:
We are close the a major release, currently in `beta` `npm i -D extract-text-webpack-plugin` and therefore I will close this PR here, feel free to open a new PR with this changes, when #540 landed on master

PR #112, opened 2015-10-05
Support loaders specified as an array

CtrlZvi:
Current if the loaders are specified as an array, then extract will
assume they are the options. If the `notExtractLoader` is specified as
an array then there will be an error when trying to split it.

It might be best to only support this change if `loaders:` is used
instead of `loader:` but even solving that problem at all seems to be
a larger issue. This at least allows for using an array of loaders
even if it still must be in a `loader:` context.

### DIFF
--- a/index.js
+++ b/index.js
@@ -127,19 +127,19 @@ ExtractTextPlugin.loader = function(options) {
 };
 
 ExtractTextPlugin.extract = function(before, loader, options) {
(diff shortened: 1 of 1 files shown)


rayshan:
+1, it's nicer to have multiple loader objects with their own query objects than to have long chained querystrings.

Issue #131, opened 2015-11-26
Module build failed: TypeError: text.forEach is not a function


Oleksandr-Silin:
I have webpack config
```javascript
    module: {
        loaders: [
            ...
            {
                test: /\.css$/,
                loader: ExtractTextPlugin.extract("style-loader", "css-loader")
            },
            ...
        ]
    },
    plugins: [
        ...
        new ExtractTextPlugin("[name].css", {
            disable: false,
            allChunks: true
        })
    ]
```
and app.js file:
```javascript
var React = require('react');
require('../css/colors.css');
require('../css/main.css');
require('../css/tags.css');
Router.run(function (Handler, state) {
    React.render(<Handler/>, document.getElementById('my-app'));
});
```
with such configuration i receive error for each css file in console output:
*Module build failed: TypeError: text.forEach is not a function
    at Object.<anonymous> (/home/src/app/node_modules/extract-text-webpack-plugin/loader.js:100:11)
    at Compiler.<anonymous> (/home/src/app/node_modules/webpack/lib/Compiler.js:214:10)
    at /home/src/app/node_modules/webpack/lib/Compiler.js:403:12....*

If I remove configuration, styles are inlined into head of html file.
```javascript
{
 disable: false,
 allChunks: true
}
```

How should i update or fix it?

sheerun:
I have the same issue

em-yuanbo:
i get the same error, and i found out that the text was not array or string but an object like ` { root: 'styles__root___2bUDW' }`.

and i just change this line https://github.com/webpack/extract-text-webpack-plugin/blob/master/loader.js#L98  from `if(typeof text === "string"){` to  `if(typeof text === "string" || !Array.isArray(text)){`  and the error gone.

but i don't know whether it is correct or not.

GreenGremlin:
Adding css-loader to the ignore option fixed this, for me.

``` JavaScript
    new SplitByPathPlugin(
      [{ name: 'vendor', path: __dirname + '/node_modules' }],
      { ignore: [__dirname + '/node_modules/css-loader'] }
    ),
```

Oleksandr-Silin:
@GreenGremlin Thank you! It helped me.

nadavsinai:
thanks @GreenGremlin !

sloot14:
Thank you! @GreenGremlin

jknight12882:
[...]

PR #134, opened 2015-12-03
fix: text.forEach is not a function

Ty3uK:
[Fixed with code written by Cleod9](https://github.com/webpack/extract-text-webpack-plugin/issues/110#issuecomment-145687012)

### DIFF
--- a/loader.js
+++ b/loader.js
@@ -112,6 +112,8 @@ module.exports.pitch = function(request) {
 				var text = this.exec(source, request);
 				if(typeof text === "string")
 					text = [[0, text]];
+				if (typeof text.forEach !== "function")
+					text = [];
 				text.forEach(function(item) {
 					var id = item[0];
 					compilation.modules.forEach(function(module) {


GreenGremlin:
For me, this was happening because I was using `webpack-split-by-path` and it was pulling css-loader out of the entry chunk. I had to tell webpack-split-by-path to ignore css-loader.

See my fix here: [Module build failed: TypeError: text.forEach is not a function #131](https://github.com/webpack/extract-text-webpack-plugin/issues/131).

Ty3uK:
I was using split-by-name-webpack-plugin and with my config webpack pulling to another chunk only angular and it's libraries.

bebraw:
Can you rebase? This could really use a test.

bebraw:
Any chance for a test?

Ty3uK:
Sorry, but i don't understand how i can test empty object into text variable and pass it through Code Coverage. Never worked with this stuff.

michael-ciniawsky:
@Ty3uK Please add a comment to the code,  so I in case this causes any regressions I can revert it :) Could you at least manually test it a bit and report ?

michael-ciniawsky:
We are getting close to a new major release and therefore I will close this for now, feel free to reopen a new PR, if you still get this regression with `npm i -D extract-text-webpack-plugin@beta`

PR #135, opened 2015-12-04
Add example for multiple plugin instances

renarsvilnis:
Added a example use case of multiple plugin instances in `README.md`.


sokra:
Thanks

Issue #115, opened 2015-10-10
Error when using with MinChunkSizePlugin


Anahkiasen:
I'm not sure why, when I enable the MinChunkSizePlugin I get this error:

```
TypeError: Cannot read property 'initial' of undefined
    at ExtractTextPlugin.<anonymous> (/Users/anahkiasen/Sites/madewithlove/madewithlove-webpack-config/node_modules/extract-text-webpack-plugin/index.js:22:8)
    at Array.forEach (native)
    at ExtractTextPlugin.mergeNonInitialChunks (/Users/anahkiasen/Sites/madewithlove/madewithlove-webpack-config/node_modules/extract-text-webpack-plugin/index.js:21:16)
    at ExtractTextPlugin.<anonymous> (/Users/anahkiasen/Sites/madewithlove/madewithlove-webpack-config/node_modules/extract-text-webpack-plugin/index.js:276:12)
    at Array.forEach (native)
    at ExtractTextPlugin.<anonymous> (/Users/anahkiasen/Sites/madewithlove/madewithlove-webpack-config/node_modules/extract-text-webpack-plugin/index.js:274:21)
    at /Users/anahkiasen/Sites/madewithlove/madewithlove-webpack-config/node_modules/async/lib/async.js:52:16
    at done (/Users/anahkiasen/Sites/madewithlove/madewithlove-webpack-config/node_modules/async/lib/async.js:248:21)
    at /Users/anahkiasen/Sites/madewithlove/madewithlove-webpack-config/node_modules/async/lib/async.js:44:16
    at /Users/anahkiasen/Sites/madewithlove/madewithlove-webpack-config/node_modules/extract-text-webpack-plugin/index.js:270:6
    at /Users/anahkiasen/Sites/madewithlove/madewithlove-webpack-config/node_modules/async/lib/async.js:52:16
    at done (/Users/anahkiasen/Sites/madewithlove/madewithlove-webpack-config/node_modules/async/lib/async.js:248:21)
    at /Users/anahkiasen/Sites/madewithlove/madewithlove-webpack-config/node_modules/async/lib/async.js:44:16
    at /Users/anahkiasen/Sites/madewithlove/madewithlove-webpack-config/node_modules/extract-text-webpack-plugin/index.js:267:13
    at /Users/anahkiasen/Sites/madewithlove/madewithlove-webpack-config/node_modules/async/lib/async.js:187:20
[...]

Issue #82, opened 2015-07-04
Getting error when use from webpack-dev-server node's api


let4be:
http://hastebin.com/igaponazaw.pl

https://github.com/webpack/extract-text-webpack-plugin/blob/master/index.js#L22

changing condition to
`if(!c || c.initial) return;`
seems to solve to problem

erwingaitano:
I'm having the same issue. Every time webpack recompiles the second time I got the error:

 `Cannot read property 'initial' of undefined`

Due to `c` being undefined.

RabidFX:
@erwingaitano This seems related to #115  indeed. Can you give a try to my fix? (Pull request [here](https://github.com/webpack/extract-text-webpack-plugin/pull/170)) I'll add some information to the PR if it solves your issue aswell.

erwingaitano:
Thanks @RabidFX !
Your solution is indeed working.

RabidFX:
@erwingaitano  Thank you for the information, I'll update the PR description accordingly.

PR #170, opened 2016-04-08
Fix proposal for issues #82 and #115. Adding extra tests for chunks existence in lists solved the problem.

RabidFX:
I am not 100 % sure that it cannot have drawbacks in some edge cases.


RabidFX:
I'll work on a test case.

brresnic:
@RabidFX, could you elaborate on why your fix fails some test cases?

RabidFX:
@brresnic Actually, it passes all of them. The two "failed checks" are not test cases, it's compliance to codecov metrics:
- codecov/patch: Only 67.00% of the lines I modified are verified by a test case (target 90.00%)
- codecov/project: Overall code coverage of the project is lower after my commits.

I'm currently writing a new test case that should address both points. (I'll try to commit it tonight.)

RabidFX:
The test case is proving to be much more challenging than I thought. It takes very specific conditions to trigger the "ghost chunk" problem, therefore covering the new code lines with a test.

The only test case I have so far is a whole project, very far from the low-footprint test cases, and with several new dependencies.

RabidFX:
The test case is done. I have removed the safeguard on addParent() since it seems I was not able to make the chunk search fail in that loop. Perhaps it's completely failproof, perhaps it's just a question of finding an "edgier" case. Hence removing the fix, to ensure code coverage.

In any case, I left only the safeguard on the addChunk() method, and the fix still works. 

@sokra, feel free to pull this in the main repo. I have no plans on working on this any further unless someone asks :smile:  
I don't know what codecov has been drinking, but with 100% patch coverage, I fail to see how project coverage would diminish, so I'm quite sure you can pull without fear. :smile:

PR #176, opened 2016-04-20
Fixed typo - nothing major :)

jonva:


bebraw:
Looks like this has been fixed so probably safe to close.

PR #157, opened 2016-03-02
Update webpack dependency to support 2.1.0-beta

jamiehodge:
Mimicking the babel-loader package.json, update the webpack peer dependency to support the 2.1.0-beta.

PR #189, opened 2016-05-20
allowing webpack 2.1.0-beta as a peerdep

antialias:
All specs pass when I use webpack 2.1.0-beta.


antialias:
bump

harounb:
#157 Appears to be the same

antialias:
Indeed. Closing in deference to https://github.com/webpack/extract-text-webpack-plugin/pull/157

PR #202, opened 2016-06-29
Fix "Cannot read property 'initial' of undefined".

stephenzhao:
when the chunks is null,  it will throw the error.

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,7 +19,7 @@ ExtractTextPlugin.prototype.mergeNonInitialChunks = function(chunk, intoChunk, c
 	if(!intoChunk) {
 		checkedChunks = [];
 		chunk.chunks.forEach(function(c) {
-			if(c.initial) return;
+			if(!c || c.initial) return;
 			this.mergeNonInitialChunks(c, chunk, checkedChunks);
 		}, this);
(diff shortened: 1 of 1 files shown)


sokra:
There seem to be some error somewhere. `chunk.chunks` should not have null elements. I'll close this PR because this doesn't fix the origin bug, but I'm interested to find the origin bug.

scottyeck:
Was this ever addressed? Would love to try and fix.

evilebottnawi:
@scottdj92  can your create minimal reproducible test repo?

r1b:
my repo has this issue https://github.com/computer-lab/salon94design.com/releases/tag/chunks-bug

r1b:
Introduced in https://github.com/computer-lab/salon94design.com/commit/52b08c5123d81532d0a7be47805ed7598710d5c0

r1b:
This commit introduces the issue: https://github.com/computer-lab/salon94design.com/commit/d7ca2cc9a7b60fb9d0ce56044302284add368d90

revolunet:
added a fix for webpack 1 too.

what about warning the user that its chunk is empty or invalid ?

Issue #215, opened 2016-07-18
Cannot resolve module [object Object]


raphaklaus:
Hi there! I'm trying to execute a webpack with extract-text, but I'm getting "Cannot resolve module '[object Object]'".

Here is a fragment.
```
    loaders: [{
      test: /\.css$/,
      loader: ExtractTextPlugin.extract({
        notExtractLoader: "style-loader",
        loader: "css-loader"
      })
    }]
```

Version of this plugin: 1.0.1
Node: 6.2.0. Tested in 4.4.0. And the same.

raphaklaus:
Well, for some reason version 1.0.1 doesn't works with webpack 1.x
I've upgraded to extract-text-webpack-plugin@v2.0.0-beta.2 and webpack@v2.1.0-beta.19 and worked perfectly.

Eschon:
Hi I'm having the same issue. I also use version 1.0.1 and my node version is 6.2.2.
Using v2.0.0-beta.2 or v2.0.0-beta.3 won't work for me because of #214 

Here's the loader config for the plugin:
```
{
  test: /\.css$/,
  loader: ExtractTextPlugin.extract({
    fallbackLoader: 'style-loader',
    loader: 'css-loader!postcss-loader'
  })
}
```

Please let me know if I can supply any additional information for resolving this issue.

Eschon:
I managed to fix it by changing my loader config to
```
{
  test: /\.css$/,
  loader: ExtractTextPlugin.extract('style', 'css-loader!postcss-loader')
}
```

Still I think there is an issue that should be resolved, because my initial config was pretty similar to what you have in the README.md

Kovensky:
You have to upgrade both `extract-text-webpack-plugin` to `^v2.0.0-beta.3` and `webpack` to `^v2.1.0-beta.19` to fix #214.

Eschon:
@Kovensky I know that #214 would have been fixed if I switched to the newest webpack beta but I don't like using beta versions of packages in production if I can avoid it.

And `ExtractTextPlugin.extract('style', 'css-loader!postcss-loader')` works without problems in `extract-text-webpack-plugin@1.0.1`

narek11:
@Kovensky doesn't `ExtractTextPlugin.extract` accepts only a single config argument ?

Kovensky:
@narek11 yes, as of the version 2 beta, and it complains at you if you do give more than one argument. Previous versions accept (up to) 3 arguments.

[...]

PR #220, opened 2016-07-28
Link to v1's documentation from v2's README

Kovensky:
Should help reduce confusion for users that are using the current stable version (c.f. #215).


sokra:
Thanks

PR #223, opened 2016-08-01
Be more specific on the peerDependencies

Kovensky:
This is not actually API-compatible with any webpack beta older than 2.1.0-beta.19.


sokra:
Thanks

PR #232, opened 2016-08-17
Fix handling of loaders as objects without a query prop

insin:
If you provide a loader as an object *without* query config, the entire object currently gets returned from `getLoaderWithQuery()` and ends up being stringified as `[object Object]` in the final loader string generated.


sokra:
Thanks

Issue #218, opened 2016-07-21
[2.0 beta] Errors thrown from loader-runner not caught/handled


oh-ren:
**Bug report**
2.0.0-beta.3 w/ wp 2.1.0-beta.20

Issue arised after upgrading to `wp` > `2.1.0-beta.15` & `extract-text-webpack-plugin@beta`.

**Current behavior:**
Errors thrown from `loader-runner` do need seem to be caught, which causes `webpack -w` to exit. They do (still) get caught when **not** using the plugin. Example:
> [...]\node_modules\loader-runner\lib\LoaderRunner.js:90
>                        throw e;
>                        ^
>
> TypeError: Cannot read property '[...]\node_modules\extract-text-webpack-plugin' of null 
> [...]

**Expected/desired behavior:**
Error message (`"Can't resolve..."`) in red, without exiting the watch process (like before).

**Reproduce**
Point to an invalid path in an `url(...)` rule.

valery-vitko:
Not quite sure if this could help. Stumbled upon this exact issue and had knocked off a bit smelly [workaround](https://github.com/Steinpilz/extract-text-webpack-plugin/commit/2753e84b2f570b782fb4d31ec9830afeba6a61b9) (`npm i -D git+ssh://git@github.com/Steinpilz/extract-text-webpack-plugin.git#type-error-investigations`).

oh-ren:
Cool, will check it out, thanks! (Though I expect will be (properly) fixed in codebase soon anyways.. :))

derpoho:
+ 1 here with beta.3...

Is there any implemented workaround yet?

oh-ren:
@valery-vitko's workaround seems to work for me.

kesne:
This seems related to #211.

Zyphrax:
@valery-vitko thank you for the workaround! It was driving me crazy that the dev-server would stop running because of my typos.

The npm command you supplied didn't work for me, this did (in case anyone else runs into that):
`npm i -D github:Steinpilz/extract-text-webpack-plugin.git#type-error-investigations`

Zyphrax:
Could you please send this in as a pull request?
That would be great!

valery-vitko:
@Zyphrax, yes, no problem - #242

Zyphrax:
Great, thank you!

oh-ren:
Closing. This seems to be no issue anymore in latest beta.

PR #242, opened 2016-09-05
fix: errors thrown from loader-runner not caught/handled

valery-vitko:
Not quite sure it should go as is, but could be used as a base for the fix.


index.js
```diff
@@ -254,14 +254,17 @@ ExtractTextPlugin.prototype.apply = function(compiler) {
 									compilation.errors.push(err);
```
Kovensky: The actual problem is that `compilation.rebuildModule` does not call the callback with an `err` even if there was a compilation failure -- when the `module.meta[NS]` is unset bug is hit, `module.errors` is (AFAICT) always non-empty.

Zyphrax:
It would be so great to have this fixed. It is the only thing preventing me from using the latest version.

bebraw:
Can you sign the CLA and rebase?

michael-ciniawsky:
Fixes #218

michael-ciniawsky:
@valery-vitko Please close and reopen the PR to trigger the CLA Bot again and sign the CLA, then rebase against current master 😛

index.js
```diff
@@ -276,15 +276,18 @@ ExtractTextPlugin.prototype.apply = function(compiler) {
...
 									return callback();
 								}
-								meta = module[NS];
-								// Error out if content is not an array and is not null
-								if(!Array.isArray(meta.content) && meta.content != null) {
-									err = new Error(module.identifier() + " doesn't export content");
-									compilation.errors.push(err);
-									return callback();
+								// TODO: Investigate "TypeError: Cannot read property '...\node_modules\extract-text-webpack-plugin' of null" in case of css errors and remove workaround check for module.meta[NS] below
```
michael-ciniawsky: Is this done or still outstanding ?

valery-vitko: Are you talking about TypeError investigation? No, but the error this workaround aims hasn't been popping out anymore for quite a while now.

michael-ciniawsky: hmm... 😛 we would still need a test for this I guess, if you don't have this regression anymore (not urgent for you), mabe better close the PR ?

valery-vitko: Yes, it don't bugs me, but it seems it's gone for good for everyone. Thus not quite sure how is it better to test this case now, though I'm not an expert in this stuff.
So, closing this PR then.

PR #87, opened 2015-07-17
fix: remove loader caching

aickin:
… a module is used in multiple places.

Issue #42, opened 2014-11-15
Unexpected: "ERROR in <blah> doesn't export content"


roblg:
When running `./node_modules/webpack/bin/webpack.js --watch`, an error shows up in the terminal after changing a CSS file ("ERROR in <blah> doesn't export content"), and the webpack compilation seems to run twice.

The high-level setup is as follows: two entry points, one (`OneEntry.js`) does a synchronous require of `Dep.js`, and the other (`TwoEntry.js`) does a require.ensure of `Dep.js`. `Dep.js` requires `Dep.css`. Changing `Dep.css` while running with `--watch` triggers the error message.

The final output doesn't seem to be affected (at least in my simple example here), but it's still a scary message. I started to step through it with node-debug, but didn't get much farther than figuring out where the error was coming from -- line 156 of `extract-text-webpack-plugin/index.js`. I'll probably look at it at little further tomorrow. 

**My webpack.config.js:**

``` javascript
var ExtractTextPlugin = require("extract-text-webpack-plugin");

module.exports = {
    target: "web",
    debug: true,
    // devtool: "source-map",
    entry: {
        One: './OneEntry.js',
        Two: './TwoEntry.js'
    },
    output: {
        path: "./target",
        filename: "[name].bundle.js"
    },
    module: {
        loaders: [            
            { test: /\.css$/, loader: ExtractTextPlugin.extract("style-loader", "css-loader") },
        ]
    },
    plugins: [
        new ExtractTextPlugin("[name].styles.css")
    ]
};
```

**OneEntry.js**

``` javascript
require('./Dep.js');

module.exports = function () {
    console.log("OneEntry");
}
```

**TwoEntry.js**

``` javascript
module.exports = function () {
    require.ensure([], function (require) {
        var One = require('./Dep.js');
    });
}
```

**Dep.js**

``` javascript
require('./Dep.css');

module.exports = function helloWorld() {
    console.log("Hello, World!");
}
```

**Dep.css**

``` css
body {
    font-weight: normal;
}
```

Console output:

```
[10:30:58] extract-text-plugin-bug-repro$ ./node_modules/webpack/bin/webpack.js --watch
Hash: 1af67a17d8cdba702d4f
Version: webpack 1.4.13
Time: 123ms
         Asset  Size  Chunks             Chunk Names
 Two.bundle.js  3850       0  [emitted]  Two
 1.1.bundle.js   332       1  [emitted]
[...]

PR #169, opened 2016-03-29
feat: allow disabling of loader cache

thebigredgeek:

PR #253, opened 2016-09-22
feat: add option to disable caching (webpack v1)

missyjcat:
A different approach to #87  which allows an isCacheable boolean option to be passed into either the ExtractTestPlugin initialization or the extract method.

re: #42 

We're working with webpack 1 so submitting this fix for webpack-1 branch.


missyjcat:
@sokra what do you think of this approach?

michael-ciniawsky:
Superseded by either #169 #87. Thx

PR #260, opened 2016-10-03
Add info about order of rules in README.md

jhnns:
The current README does not tell in which order the generated text will appear in the output file. Since the order of rules is important in CSS, we should add a note on this. It is also important that this behavior is currently based implicitly on the way webpack is ordering the modules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ module.exports = {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


TheLarkInn:
Lgtm

jhnns:
Merge? 😁

PR #262, opened 2016-10-05
Fix and improve code examples in README

jhnns:
Reflects the latest changes to the loader configuration


szimek:
According to https://webpack.js.org/how-to/upgrade-from-webpack-1/ (is it up-to-date?) there are some more changes that could be shown in the README file, e.g. `module: { loaders: ... }` is now `module: { rules: ... }`, `query` is now `options` etc.

jhnns:
Yes, but then we need to change the implementation first :). See #265 
This should be merged sooner.

PR #266, opened 2016-10-07
Pass proper loader `options` to children.

izaakschroeder:
Using just the serialized request means that query objects (that is to say, those which are not strings), get munged if they have anything non-serializable in them. This means passing data to `postcss` via `query` is impossible. To remedy this, we detect when the original request is being processed and apply the original loader transformations instead of whatever came through the serialized version.

Build seems to be _only_ failing on `node@0.12` but `webpack@2` doesn't support that.. so..

See:
- https://github.com/postcss/postcss-loader/pull/104
- https://github.com/webpack-config/webpack-config-postcss/pull/10

/cc @sokra

### DIFF
--- a/loader.js
+++ b/loader.js
@@ -20,6 +20,7 @@ module.exports = function(source) {
 module.exports.pitch = function(request) {
 	if(this.cacheable) this.cacheable();
 	var query = loaderUtils.parseQuery(this.query);
+	var loaders = this.loaders.slice(this.loaderIndex + 1);
 	this.addDependency(this.resourcePath);
 	// We already in child compiler, return empty bundle
 	if(this[NS] === undefined) {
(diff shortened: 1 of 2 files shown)


bebraw:
Looks like a good idea to me. I wouldn't worry about Node 0.12.

michael-ciniawsky:
@bebraw @d3viant0ne Ahh... damn 😛 I missed that one, it is actually quite important for postcss-loader just saw it the popping up in timeline.

@izaakschroeder  Could you reword the commits to e.g

`fix(loader): pass proper loader 'options' to children` 
`chore(package): update loader-utils v0.2.3...0.2.16`

 🙏  so we can make a quick release ?

d3viant0ne:
@michael-ciniawsky - I'll see to it in a minute.

d3viant0ne:
Included in `v2.0.0-rc.1`

PR #269, opened 2016-10-15
Filename filtering

lcxfs1991:
change the filename for css

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ new ExtractTextPlugin([id: string], filename: string, [options])
   * `allChunks` extract from all additional chunks too (by default it extracts only from the initial chunk(s))
   * `disable` disables the plugin
   * `filenamefilter` function to modify path and filename before files are emitted
+    
     ```
     {
       filenamefilter: function(filename) {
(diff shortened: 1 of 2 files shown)


bebraw:
The idea looks good to me. I'm not sure of the naming so maybe others can comment on that.

Please sign the CLA in any case.

lcxfs1991:
@bebraw have you also considered this for the plugin for webpack2?

bebraw:
It's up to Tobias.

I think I would overload `filename` now rather than having another field. Then you would end up with:

```javascript
new ExtractTextPlugin(({ name }) => `stylesheets/${name}.css`);
new ExtractTextPlugin({ filename: ({ name }) => `stylesheets/${name}.css` });
```

That's as flexible as it can get. The same idea would work elsewhere in webpack too if this scheme is acceptable.

lcxfs1991:
yep. Shall we wait for the scheme, maybe next version of webpack?

bebraw:
It needs a design decision, yeah.

bebraw:
@lcxfs1991 Would it be possible for you to refactor the idea to follow the specification I discussed? Instead of a new field we would overload the current `filename`. That would give us the feature without growing the API.

lcxfs1991:
@bebraw I have refractor my code and add test cases. Please check.

https://github.com/webpack-contrib/extract-text-webpack-plugin/pull/423

lcxfs1991:
this one can be closed

PR #276, opened 2016-10-29
allow both `loader.options` and `loader.query`, add test cases

dlindenkreuz:
Webpack v2 recommends using the `options` key instead of `query`: https://webpack.js.org/configuration/module/#useentry

See also: https://github.com/webpack/extract-text-webpack-plugin/commit/b79da9f375f02c349cb51cdff36ddb3c25974b69#commitcomment-19621830

### DIFF
--- a/index.js
+++ b/index.js
@@ -131,8 +131,9 @@ module.exports = ExtractTextPlugin;
 // modified from webpack/lib/LoadersList.js
 function getLoaderWithQuery(loader) {
 	if(isString(loader)) return loader;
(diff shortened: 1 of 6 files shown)


bebraw:
LGTM. I think that Node 0.12 check can be discarded given webpack 2 is Node 4+ now.

d3viant0ne:
@dlindenkreuz - This should no longer be an issue as of https://github.com/webpack-contrib/extract-text-webpack-plugin/releases/tag/v2.0.0-rc.0

PR #295, opened 2016-11-29
fix: usage with local classnames

joeybaker:
Prior to this, when used with css-loader and `modules: true` the compiled
class names would not get formatted properly into the JS since
`resultSource` was always `undefined`. This sets `resultSource` to an empty
string (which seems to have no detrimental effects).

### DIFF
--- a/test/TestCases.test.js
+++ b/test/TestCases.test.js
@@ -29,9 +29,9 @@ describe("TestCases", function() {
 			webpack(options, function(err, stats) {
 				if(err) return done(err);
 				if(stats.hasErrors()) return done(new Error(stats.toString()));
-				var testFile = path.join(outputDirectory, "test.js");
+				var testFile = path.join(testDirectory, "test.js");
 				if(fs.existsSync(testFile))
-					require(testFile)(suite);
+					require(testFile)(describe);
 				var expectedDirectory = path.join(testDirectory, "expected");
 				fs.readdirSync(expectedDirectory).forEach(function(file) {
 					var filePath = path.join(expectedDirectory, file);
--- a/test/cases/style-loader-replacement/webpack.config.js
+++ b/test/cases/style-loader-replacement/webpack.config.js
@@ -1,0 +1,26 @@
+var ExtractTextPlugin = require('../../../');
+module.exports = {
+	entry: "./index",
+	module: {
+		loaders: [{
+          test: /\.css$/,
+          use: ExtractTextPlugin.extract({
+			use: {
+			  loader: 'css-loader',
+			  options: {
+				// turn on CSS Modules
+				modules: true,
+				localIdentName: 'cssmodule-[path]-[name]',
+				sourceMap: true
+			  }
+			}
+		  }),
+          exclude: /node_modules/
+        }]
+	},
+	devtool: "source-map",
+	plugins: [
+		new ExtractTextPlugin({filename: 'file.css', allChunks: true})
+	]
+};
+
--- a/loader.js
+++ b/loader.js
@@ -38,7 +38,7 @@ module.exports.pitch = function(request) {
 		if(query.remove) {
 			resultSource = "// removed by extract-text-webpack-plugin";
 		} else {
-			resultSource = undefined;
+			resultSource = '';
 		}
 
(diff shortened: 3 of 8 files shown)


test/cases/style-loader-replacement/webpack.config.js
```diff
@@ -0,0 +1,45 @@
+var ExtractTextPlugin = require("../../../");
+module.exports = {
+	entry: "./index",
+	module: {
+		loaders: [
+			// { test: /\.css$/, loader: ExtractTextPlugin.extract({
```
michael-ciniawsky: Why is this commented out? Does it require changes or could it be removed ? 🙃

test/cases/style-loader-replacement/webpack.config.js
```diff
@@ -0,0 +1,45 @@
...
+              , options: {
+                // turn on CSS Modules
+                modules: true
+                // needed to get postcss to work https://github.com/postcss/postcss-loader#css-modules
+                // , importLoaders: true
+                , localIdentName: '[path]-[name]'
+                , sourceMap: true
+              }
+            }
+            // , {
```
michael-ciniawsky: Why is this commented out? Does it require chnanges or could it be removed ?

test/cases/style-loader-replacement/webpack.config.js
```diff
@@ -0,0 +1,45 @@
...
+			// 	loader: { loader: "css-loader", query: {
+			// 		sourceMap: true
+			// 	} }
+			// }) }
+
+        {
+          test: /\.css$/,
+          use: [
+            ExtractTextPlugin.extract('css')
+            , {
```
michael-ciniawsky: Please move the `,` to the preceding line end 😛

michael-ciniawsky:
@joeybaker Could you please provide a before vs. after example in addition, just to clearify 😛

michael-ciniawsky:
@joeybaker Is this still relevant ? friendly ping :)

joeybaker:
@michael-ciniawsky I'm honestly not sure, I'm not running this in prod anymore. But I suspect this is still needed. I'm happy to cleanup the styling.

michael-ciniawsky:
Would be appreciated if you can finish it nevertheless :)

joeybaker:
@michael-ciniawsky appveyor looks like it's not configured correctly? Other than that, changes made.

test/TestCases.test.js
```diff
@@ -29,9 +29,9 @@ describe("TestCases", function() {
 			webpack(options, function(err, stats) {
 				if(err) return done(err);
 				if(stats.hasErrors()) return done(new Error(stats.toString()));
-				var testFile = path.join(outputDirectory, "test.js");
+				var testFile = path.join(testDirectory, "test.js");
```
michael-ciniawsky: What does this change exactly change do ? :)

test/TestCases.test.js
```diff
@@ -29,9 +29,9 @@ describe("TestCases", function() {
...
 				if(fs.existsSync(testFile))
-					require(testFile)(suite);
```
michael-ciniawsky: What does this change exactly change do ?

michael-ciniawsky:
Yep, don't mind `appveyor` :)

test/TestCases.test.js
michael-ciniawsky: The both look, like hanging in there somehow :D

test/TestCases.test.js
joeybaker: This fixes an apparent bug. As far as I can tell, this whole section was setup to allow us to test output js, but wasn't used up till now. Here, it's simply looking in the wrong directory for the original test file.

test/TestCases.test.js
joeybaker: Just fixes a bug. `suite` is undefined. We meant to pass in `describe`

test/TestCases.test.js
michael-ciniawsky: uuhh.. 😟. That would be really bad

joeybaker: I'm happy to be wrong 😬

michael-ciniawsky:
We are close the a major release, currently in `beta` `npm i -D extract-text-webpack-plugin` and therefore I will close this PR here. The test are updated there now. Thx :)

PR #300, opened 2016-12-05
Throw error if output filename is not defined.

### LINKED ISSUE #284: Should error on invalid ExtractTextPlugin config object
The new version of ExtractTextPlugin helpfully told me to update the arguments when calling `new ExtractTextPlugin()`, but it did not error when I provided an object with the property `fileName` instead of `filename`. This led to some unusual errors in webpack and the cause was hard to track down.

rgrwkmn:
Fixes #284

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,6 +14,17 @@ var NS = fs.realpathSync(__dirname);
 
 var nextId = 0;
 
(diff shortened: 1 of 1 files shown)


rgrwkmn:
Dangit, I messed up my git user when I created the commit and my git fu is not strong enough to fix this in place.

PR #301, opened 2016-12-05
Throw error if filename is not defined.

### LINKED ISSUE #284: Should error on invalid ExtractTextPlugin config object

rgrwkmn:
Fixes #284

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,6 +14,17 @@ var NS = fs.realpathSync(__dirname);
 
 var nextId = 0;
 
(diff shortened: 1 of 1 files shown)


bebraw:
If #358 is done right, the need for this check might go away. Please have a look.

bebraw:
The schema was merged. Let's close this.

Issue #246, opened 2016-09-15
publicPath parameter doesn't work?


yugasun:
When I use extract-text-webpack-plugin output css file into a folder, the image url is wrong, my webpack config is below:

```
"use strict";
var path = require('path');
var webpack = require('webpack');

// webpack plugins
var HtmlWebpackPlugin = require('html-webpack-plugin');
var ExtractTextPlugin = require('extract-text-webpack-plugin');


module.exports = {
    entry: path.resolve(__dirname, 'app/scripts/main.js'),
    output: {
        path: path.resolve(__dirname, "build"),
        //publicPath: path.resolve(__dirname, "build"),
        filename: 'js/[name]-[hash].js'
    },
    module: {
        loaders: [
            {
                test: /\.jsx?$/,   
                loader: 'babel',    
                exclude: /(node_modules|bower_components)/,
                query: {
                    presets: ['react', 'es2015']
                }
            },
            {
                test: /\.scss$/,
                exclude: /(node_modules|bower_components)/,
                loader: ExtractTextPlugin.extract('style', 'css!sass!postcss')
            },
            {
                test: /\.css$/,
                exclude: /(node_modules|bower_components)/,
                loader: ExtractTextPlugin.extract('style', 'css!postcss')
            },
            {
                test: /\.(woff|woff2|svg|eot|ttf)/,
                loader: 'url?prefix=font/&limit=10000'
            },
            {
                test: /\.(png|jpe?g|gif|svg)$/,
                loaders: [
                    'file?name=images/[name].[ext]',
                    //'image-webpack'
                ]
            }
        ]
    },

    // image handle
    //imageWebpackLoader: {
    //    progressive: true,
    //    optimizationLevel: 3,
    //    interlaced: false,
    //    pngquant: {quality: "65-70", speed: 4}
    //},

    postcss: [
        require('autoprefixer')
    ],

    plugins: [
        new webpack.BannerPlugin('This file is created by yugasun'),
        new HtmlWebpackPlugin({
            title: 'RedChild',
            template: __dirname + '/app/template/index.tmpl.html'
        }),
        new webpack.optimize.OccurenceOrderPlugin(),
        new webpack.optimize.UglifyJsPlugin(),
        new ExtractTextPlugin('css/[name]-[hash].css')
    ]
};
```

my output 'build' folder tree like this:

```
.
├── css
│   └── main-72c33a6c0ad73a5a0403.css
├── images
│   ├── arrow.png
[...]

PR #305, opened 2016-12-11
#cache publicPath

githoniel:
cache publicPath to fixed the bug that publicPath only take effect for the first file #246


bebraw:
@githoniel Is there some particular reason why you closed this?

githoniel:
@bebraw  it actually fixed the bugs in my webpack1 projects，but i have upgraded  to webpack2 and those project works well without it

bebraw:
Alright. Our focus is on webpack 2 so I guess it's safe to close this again. Thanks for clarification.

PR #313, opened 2016-12-19
Bump `async`.

wtgtybhertgeghgtwtg:
To match [the version](https://github.com/webpack/webpack/blob/master/package.json#L11) `webpack` uses.


sokra:
Thanks

PR #316, opened 2016-12-22
added installation to readme

JPeer264:
I added the installation guide. This might be useful for couple of people.


bebraw:
Good enough. We might want to standardize on these across projects at some point, though.

PR #343, opened 2017-01-19
Return an array of loader objects from extract()

insin:
This allows you to `use:` the output of `extract()` by not munging everything into a string (which `use` config should obviate the need for in Webpack 2).

It's far less surprising and makes it possible to [provide plugins to preprocessor loaders directly via loader `options`](https://github.com/webpack/extract-text-webpack-plugin/issues/265#issuecomment-273773975) instead of messing around with `LoaderOptionsPlugin`

This change is as per tmair@1414036 - if @tmair wants to create their own PR for it, I'm happy to close this one.

### DIFF
--- a/index.js
+++ b/index.js
@@ -148,7 +148,7 @@ function isString(a) {
 }
 
 ExtractTextPlugin.loader = function(options) {
-	return { loader: require.resolve("./loader"), query: options };
+	return { loader: require.resolve("./loader"), options: options };
 };
 
 ExtractTextPlugin.prototype.applyAdditionalInformation = function(source, info) {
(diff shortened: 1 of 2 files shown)


bebraw:
You need to sign the CLA. Apart from that, big 👍 from me.

tmair:
@insin Go for it. I just did not know if my changes did break someting

tmair:
@insin Travis is failing because of an old node version. If you rebase your canges to the current master the isse should be gone (See https://github.com/webpack/extract-text-webpack-plugin/pull/335)

khankuan:
@insin thanks for the fix. Hopefully it can be rebased and merge soon. That's the main blocker I've found migrating to webpack2. Thanks!

insin:
@TheLarkInn Added some unit tests for the current implementation and tweaked it slightly to use/support `options`.

Are there other use cases which `extract()` should support?

e.g. if you specify a loader as a string with query params, is it ok that it will be now passed straight through as-is, or is that something for Webpack itself to worry about?

jakwuh:
Really waiting for this PR to be merged 👍

d3viant0ne:
Thanks @insin - This will be in `2.0.0-rc.0` which will be published in short order

PR #347, opened 2017-01-20
Adds "webpack 2.2.0" as option for peer dependency

### LINKED ISSUE #346: Unable to install in project with webpack 2.2.0
**package.json**

```
{
  "name": "hello",
  "description": "NA",
  "version": "1.0.0",
  "author": " <>",
  "private": true,
  "scripts": {},
  "dependencies": {},
  "devDependencies": {
    "babel-core": "^6.0.0",
    "babel-loader": "^6.0.0",
    "babel-preset-es2015": "^6.0.0",
    "css-loader": "^0.25.0",
    "file-loader": "^0.9.0",
    "webpack": "^2.2.0"
  }
}
```

Everything install fine there. Then, when trying to add `extract-text-webpack-plugin`, I get this error:

```
npm install --save-dev extract-text-webpack-plugin
hello@1.0.0 /Users/example/hello
├─┬ extract-text-webpack-plugin@1.0.1 
│ └── async@1.5.2 
└── UNMET PEER DEPENDENCY webpack@2.2.0

npm WARN extract-text-webpack-plugin@1.0.1 requires a peer of webpack@^1.9.11 but none was installed.
```

romellem:
Resolves #346

Not sure if we can remove `^2.2.0-rc.0` from the `peerDependencies` altogether, but this does fix the related issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "author": "Tobias Koppers @sokra",
   "description": "Extract text from bundle into a file.",
   "peerDependencies": {
-    "webpack": "^2.1.0-beta.19 || ^2.2.0-rc.0"
+    "webpack": "^2.1.0-beta.19 || ^2.2.0-rc.0 || 2.2.0"
   },
   "dependencies": {
     "async": "^2.1.2",


taion:
This seems strictly wrong. `^2.2.0-rc.0` matches `v2.2.0`, but `2.2.0` wouldn't even match, say, `v2.2.1`.

romellem:
> This seems strictly wrong. ^2.2.0-rc.0 matches v2.2.0, but 2.2.0 wouldn't even match, say, v2.2.1.

You are correct, I meant to put in `^2.2.0`, but in either case, I was just looking for a solution to the original issue I was seeing. However, it turns out there is already a [work-around](https://github.com/webpack/extract-text-webpack-plugin/issues/346#issuecomment-274116917), although it isn't considered "stable"

> Go through `@beta` - `npm install --save-dev extract-text-webpack-plugin@beta`.

PR #358, opened 2017-01-25
Update to include JSON schema and some validation

scottdj92:


schema/valid.js
```diff
@@ -0,0 +1,14 @@
...
+var json = require('./schema.json');
+
+module.exports = function validate(data) {
+	var validSchema = ajv.compile(json);
+	var valid = validSchema(data);
+
+	if(!valid) {
+		throw new Error("Your ExtractTextPlugin config is not correct. Please double check.");
+	} else {
+		console.log('your stuff works');
```
michael-ciniawsky: Is this printed on success every time ? 😛 I like the idea of a positive signal but maybe something like `PLUGIN_NAME:  ✅ `

bebraw: Yeah. Better just to be silent by default.

schema/schema.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
```
bebraw: You should probably set `"additionalProperties": false,` on root level as that gives stricter validation.

d3viant0ne:
@scottdj92 - A couple of things before this can be landed.

1.) Need rebase to pick up some housekeeping in master
2.) It's currently failing on Travis, this will need to be addressed ( see review comment above )
3.) This could also use at least one test & it's negative counterpart

bebraw:
Note that ajv can deal with default values. That could be another PR, though.

schema/valid.js
```diff
@@ -8,7 +8,5 @@ module.exports = function validate(data) {
 
 	if(!valid) {
 		throw new Error("Your ExtractTextPlugin config is not correct. Please double check.");
```
bebraw: You can get a string message out of ajv. See https://github.com/webpack-contrib/purifycss-webpack for the basic idea. I do also defaults there.

scottdj92:
Regarding adding tests, would it only pass testing if the value returned by ajv is `true`, meaning the passed in object matches the schema and vice versa? Or would the test have to go deeper than that?

bebraw:
You can test schema success/failure at minimum, yeah.

scottdj92:
Seeing [#1465](https://github.com/webpack/webpack/issues/1465) when running `npm test`. Not sure if webpack-core has been updated. Currently every other test passes, except for this one error.

schema/validator.js
```diff
@@ -0,0 +1,12 @@
+var Ajv = require('ajv');
+var ajv = new Ajv({allErrors: true});
+var json = require('./schema.json');
+
+module.exports = function validate(data) {
+	var validSchema = ajv.compile(json);
+	var valid = validSchema(data);
+
+	if(!valid) {
+		throw new Error("Your ExtractTextPlugin config is not correct. Please double check.");
```
bebraw: You can get possible error text through `ajv.errorsText()`. Better throw that instead.

scottdj92: Added :D

scottdj92:
All tests now pass as of 1/26/2017.

PR #374, opened 2017-01-29
fix - Implement custom `toString` for `extract`

### LINKED ISSUE #364: ERROR in ./~/bootstrap-loader/lib/bootstrap.loader.js?extractStyles!./~/bootstrap-loader/no-op.js
After updating to rc0 from 2.0.0-beta.5:

ERROR in ./~/bootstrap-loader/lib/bootstrap.loader.js?extractStyles!./~/bootstrap-loader/no-op.js
Module not found: Error: Can't resolve '[object Object],[object Object],[object Object],[object Object],[object Object],[object Object]./lib\bootstrap.styles.loader.js' in 'C:\PROJECTS\UCB_IMPRESTO_FE\node_modules\bootstrap-loader'
 @ ./~/bootstrap-loader/lib/bootstrap.loader.js?extractStyles!./~/bootstrap-loader/no-op.js 1:21-594
 @ ./~/bootstrap-loader/extractStyles.js
 @ ./src/vendors.ts

### LINKED ISSUE #370: Fail hard if `loader: plugin.extract` is used
This is an important check to make as `extract` works only with `use`.

bebraw:
This should fix the behavior with loaders that rely on pitching. Since it needs to deal with a string, `extract` should return the loader in a query format as well.

The change allows the `extract` to be used with `loader` field again.

There's one gotcha - the current solution doesn't handle functions (postcss) in any special way. That may require further thinking.

Closes #364.
Closes #370.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@
 	Author Tobias Koppers @sokra
 */
 var fs = require('fs');
+var querystring = require('querystring');
 var ConcatSource = require("webpack-sources").ConcatSource;
 var async = require("async");
 var ExtractedModule = require("./ExtractedModule");
@@ -202,9 +203,17 @@ ExtractTextPlugin.prototype.extract = function(options) {
 	options = mergeOptions({omit: before.length, remove: true}, options);
 	delete options.loader;
 	delete options.fallbackLoader;
-	return [this.loader(options)]
+	var result = [this.loader(options)]
 		.concat(before, loader)
 		.map(getLoaderObject);
+
+	result.toString = function() {
+		return result.map(function(o) {
+			return o.loader + '?' + querystring.stringify(o.options);
+		}).join('!');
+	};
+
+	return result;
 }
 
 ExtractTextPlugin.extract = ExtractTextPlugin.prototype.extract.bind(ExtractTextPlugin);
--- a/test/extract.test.js
+++ b/test/extract.test.js
@@ -37,133 +37,230 @@ describe("ExtractTextPlugin.extract()", function() {
 
 	context("specifying loader", function() {
 		it("accepts a loader string", function() {
-			ExtractTextPlugin.extract("css-loader").should.deepEqual([
-				{ loader: loaderPath, options: { omit: 0, remove:true } },
-				{ loader: "css-loader" }
-			]);
+			var result = ExtractTextPlugin.extract("css-loader");
+
+			delete result.toString;
+
+			should.deepEqual(
+				result,
+				[
+					{ loader: loaderPath, options: { omit: 0, remove:true } },
+					{ loader: "css-loader" }
+				]
+			);
 		});
 
 		it("accepts a chained loader string", function() {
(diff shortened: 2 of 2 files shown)


scottdj92:
Looks ok to me. I'd like to hear @sokra's opinion about how to handle functions before merging, though.

bebraw:
Yup. Same here. It's related to `ident` and its handling I think. Maybe the loader should be able to hook into the core mechanism of generating one somehow. Then it could dig the function to execute from the serialized result.

index.js
```diff
@@ -202,9 +203,17 @@ ExtractTextPlugin.prototype.extract = function(options) {
 	options = mergeOptions({omit: before.length, remove: true}, options);
 	delete options.loader;
 	delete options.fallbackLoader;
-	return [this.loader(options)]
+	var ret = [this.loader(options)]
```
Arturszott: In one place you write 'result' and there 'ret'. I think that first one if more verbose option 👍

bebraw: Ah, yeah. That's a habit. I'll get this fixed up. Thanks.

scottdj92:
Could we check the passed in argument for `extract` instead and then add flow control? As far as I can see this current PR would solve a problem where a string is passed in. I'd think that we could check to see whether or not it is a function and change the plugin to behave accordingly.

However, I'm not 100% sure on what exactly the functions return though. PostCSS may return a full string for this plugin to extract.

bebraw:
@scottdj92 It's used like this at the font-awesome-loader:

```javascript
  var styleLoaderCommand = 'require(' + JSON.stringify('-!' + styleLoader + '!' +
      require.resolve('./font-awesome-styles.loader.js') + '!' + configFilePath) + ');';
```

There's not enough context info so I had to go through `toString` to adapt to this situation.

sokra:
I don't think this is necessary. This should be fixed on font-awesome-loader level.

`extract` returns an array. You should pass this array to `use`. It's not a string! Don't use it as string! String won't work for complex options in extracted loader.

bebraw:
> extract returns an array. You should pass this array to use. It's not a string! Don't use it as string! String won't work for complex options in extracted loader.

`use` gets an array of objects right now but `loader` won't work with an array of objects. In order to maintain compatibility with both, we would have to find a structure that is compatible with both while `toString` isn't used at all.

That `toString` was written specifically font-awesome-loader pitch phase in mind as it seems to rely on a query string. But it's true that loader could be rewritten.

bebraw:
Ok, I see now. `use: {loader: Array}`, `loader: {loader: Array}` just might do the trick. So if we wrap the current output like that and drop `toString`, this could work.

bebraw:
Apparently the current implementation actually works against `loader` after all since it can deal with the array syntax due to backwards compatibility reasons. Since we agreed to skip the `toString` step, this PR can be closed.

I think we still may want to document how to fix pitch loaders depending on the old behavior (query over an array of objects).

Issue #375, opened 2017-01-29
Build failing with "Error: data['notExtractLoader'] should NOT have additional properties" - Is notExtractLoader still supported?


jgliner:
Hi, my build has been failing for the past day or two. I have tried updating to the latest version (rc.2), as well as rolling back a couple versions with no success. I did some digging and found out that `notExtractLoader` from my webpack config 

Stacktrace:
```
BASE_FOLDER/node_modules/extract-text-webpack-plugin/schema/validator.js:11
                throw new Error(ajv.errorsText());
                ^

Error: data['notExtractLoader'] should NOT have additional properties
    at validate (BASE_FOLDER/node_modules/extract-text-webpack-plugin/schema/validator.js:11:9)
    at Function.ExtractTextPlugin.extract (BASE_FOLDER/node_modules/extract-text-webpack-plugin/index.js:190:3)
    at Object.<anonymous> (BASE_FOLDER/webpack.client.prod.js:74:35)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
```

Going into the commit history, I have determined that [03bb4aa](https://github.com/webpack-contrib/extract-text-webpack-plugin/commit/03bb4aa50f4cf2b580512d066c82886230303571) is likely the culprit. Below is the plugin being used in my webpack config:
``` js
...
      {
        test: /\.css$/,
        loader: ExtractTextPlugin.extract({
          notExtractLoader: 'style-loader',
          loader: 'css-loader',
        }),
      },
...
```
Namely, your loader schema does not include `notExtractLoader`. Is this still supported? It is still referenced in the `example` folder, so I went ahead and made a PR for your consideration. Thanks in advance!

bebraw:
`loader: ExtractTextPlugin.extract` should become `use: ExtractTextPlugin.extract`. I have a fix for `loader`, but it's waiting on feedback from Tobias.

bebraw:
The validation fails because `notExtractLoader` field does not do anything. I think you may have meant `fallbackLoader`.

bebraw:
[...]

PR #376, opened 2017-01-29
Add notExtractLoader to loader-schema.json

jgliner:
See #375


bebraw:
It's `fallbackLoader` according to the docs.

PR #388, opened 2017-02-01
feat - Use `use`/`fallback` pair at `extract` over loader`/`fallbackLoader`

### LINKED ISSUE #382: Move API to use/fallback in v2
To comply with `use`, the API should look like this (to quote @sokra):

```javascript
use: ETP.extract({
  fallback: [
    { loader: "style-loader", options: {} }
  ],
  use: [
    { loader: "css-loader" },
    { loader: "postcss-loader", options: { plugin() { ... } } }
  ]
})
```

We probably should give deprecation messages for the old fields.

bebraw:
Closes #382.


example/webpack.config.js
```diff
@@ -14,8 +14,8 @@ module.exports = {
 	module: {
 		loaders: [
 			{ test: /\.css$/, use: ExtractTextPlugin.extract({
-				fallbackLoader: "style-loader",
-				loader: "css-loader?sourceMap",
+				fallback: "style-loader",
+				use: "css-loader?sourceMap",
```
SimenB: Change to use `options`?

bebraw: Yeah, I'll do that. Thanks for the heads up.

bebraw: Done.

PR #390, opened 2017-02-02
refactor: don't throw errors in case of child context

digitalkaoz:
will work with this conditional webpack rule (in conjunction with html-webpack-loader):

```js
            {
                test: /\.scss$/,
                oneOf: [
                    {test: /html-webpack-plugin/, use: "null-loader"},
                    {
                        use: ExtractTextPlugin.extract({
                            fallbackLoader: 'style-loader',
                            loader: ['css-loader', 'sass-loader']
                        })
                    }
                ]
            },
```

### DIFF
--- a/loader.js
+++ b/loader.js
@@ -23,12 +23,7 @@ module.exports.pitch = function(request) {
 	var loaders = this.loaders.slice(this.loaderIndex + 1);
 	this.addDependency(this.resourcePath);
 	// We already in child compiler, return empty bundle
-	if(this[NS] === undefined) {
-		throw new Error(
-			'"extract-text-webpack-plugin" loader is used without the corresponding plugin, ' +
-			'refer to https://github.com/webpack/extract-text-webpack-plugin for the usage example'
-		);
-	} else if(this[NS] === false) {
+	if(this[NS] === undefined || this[NS] === false) {
 		return "";
 	} else if(this[NS](null, query)) {
 		if(query.omit) {


digitalkaoz:
i signed to CLA but it wont show up here...

bebraw:
Yeah, the CLA checker is currently disabled. No need to worry about that.

I think we have to wait on @sokra on whether this is a good idea or not. It would fix your use case, but make it easier to miss the connection.

digitalkaoz:
yeah i know...but the line below returned an empty string without throwing in case of a child context...so i thought i would be ok to fail silent

bebraw:
Maybe a warning would be a softer approach. At worst case we'll add a flag for disabling the error. But it's Tobias' call to make.

michael-ciniawsky:
@digitalkaoz What is your particular usecase for this? In any case, please close and reopen the PR to trigger CLA Bot again and sign to CLA

We definitely need at least to `console.warn('Current Error Message')` if this going to land :)

loader.js
```diff
@@ -23,12 +23,7 @@ module.exports.pitch = function(request) {
...
 	this.addDependency(this.resourcePath);
 	// We already in child compiler, return empty bundle
-	if(this[NS] === undefined) {
-		throw new Error(
-			'"extract-text-webpack-plugin" loader is used without the corresponding plugin, ' +
-			'refer to https://github.com/webpack/extract-text-webpack-plugin for the usage example'
-		);
-	} else if(this[NS] === false) {
+	if(this[NS] === undefined || this[NS] === false) {
 		return "";
```
michael-ciniawsky:
```
console.warn(`
  The extract-text-webpack-plugin loader is used without the corresponding plugin,
  in case this is unintentional refer to https://github.com/webpack/extract-text-webpack-plugin 
  for correct usage (example)
`)
```

Or similiar 😛

d3viant0ne: Should probably emit here no? ( Ha, finally got to not be on the receiving end of `console.warn` hand slap! )

michael-ciniawsky: good catch, yep `this.emitWarning` 😛, example updated ☝️

Nevraeka:
Any updates on this?

michael-ciniawsky:
@digitalkaoz friendly ping

michael-ciniawsky:
We are close the a major release, currently in `beta` `npm i -D extract-text-webpack-plugin` and therefore I will close this PR here, feel free to open a new PR with this changes, when #540 landed on master

PR #407, opened 2017-02-15
beta version warning and css-loader options

pragyandas:
Added beta-version code example warning and using options with css-loader to support css-modules

1. [Read and sign the CLA](https://cla.js.foundation/webpack/webpack.js.org). This needs to be done only once. PRs that haven't signed it won't be accepted.
2. Check out the [development guide](https://webpack.js.org/development/) for the API and development guidelines.
3. Read through the PR diff carefully as sometimes this can reveal issues. The work will be reviewed, but this can save some effort.
4. Remove these instructions from your PR as they are for your eyes only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,8 @@ npm install --save-dev extract-text-webpack-plugin
 
 > :warning: For webpack v1, see [the README in the webpack-1 branch](https://github.com/webpack/extract-text-webpack-plugin/blob/webpack-1/README.md).
 
+> :warning: The following example code only works with the beta version.
+
 ```js
 const ExtractTextPlugin = require("extract-text-webpack-plugin");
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,6 +48,29 @@ module.exports = {
...
+const ExtractTextPlugin = require("extract-text-webpack-plugin");
+
+module.exports = {
+  module: {
+    rules: [
+      {
+        test: /\.css$/,
+        use: ExtractTextPlugin.extract({
+          fallback: "style-loader",
+          use: 'css-loader?modules,localIdentName="[name]-[local]-[hash:base64:6]"'
```
bebraw: It would be better to write this as

```
use: {
  loader: 'css-loader',
  options: {
    modules: true,
    localIdentName: '[name]-[local]-[hash:base64:6]'
  }
}
```

since that reads better.

pragyandas: I tried it like that but with 

> ExtractTextPlugin.extract

 it doesn't work.

bebraw: How did it fail? Sounds like a bug.

pragyandas: Should I raise an issue?

bebraw: Yup. I would love to understand how it fails. The statements should be equal.

michael-ciniawsky: @pragyandas Are you on `extract-text-webpack-plugin > = v2.0.0-rc.3` ? The options in `{Object}` notation looks nearly the same as the one I use everyday, the only diverence is that they are in an `{Array}`, bc I have other loaders in my setup. Is `use` allowed to take an `{Object}` as value?

```diff
+ use: [
  {
     loader: 'css-loader',
     options: {
       modules: true,
       localIdentName: '[name]-[local]-[hash:base64:6]'
    }
  }
+ ]
```

bebraw: `use` should work with an object. If it doesn't, that's a bug.

michael-ciniawsky: yep just tested quickly and it seems to work :)

bebraw: I guess the question is why it doesn't work for @pragyandas. I would need a project to study to say why.

michael-ciniawsky: sure 🙃

bebraw:
Not in beta anymore. Better skip. Thanks anyhow.

PR #423, opened 2017-02-21
refractor filename modify logic

lcxfs1991:
1. [Read and sign the CLA](https://cla.js.foundation/webpack/webpack.js.org). This needs to be done only once. PRs that haven't signed it won't be accepted.
2. Check out the [development guide](https://webpack.js.org/development/) for the API and development guidelines.
3. Read through the PR diff carefully as sometimes this can reveal issues. The work will be reviewed, but this can save some effort.
4. Remove these instructions from your PR as they are for your eyes only.


README.md
```diff
@@ -155,6 +155,29 @@ module.exports = {
 }
 ```
 
+### Modify filename
+
+`filename` paramter could be `Object`. It accepts `format` and `modify` callback as attributes.
```
bebraw: /s/paramter/parameter

README.md
```diff
@@ -68,7 +68,7 @@ new ExtractTextPlugin(options: filename | object)
 |Name|Type|Description|
 |:--:|:--:|:----------|
 |**`id`**|`{String}`|Unique ident for this plugin instance. (For advanced usage only, by default automatically generated)|
-|**`filename`**|`{String}`|Name of the result file. May contain `[name]`, `[id]` and `[contenthash]`|
+|**`filename`**|`{String|Object}`|Name of the result file. May contain `[name]`, `[id]` and `[contenthash]`|
```
bebraw: It could accept a function directly instead of an object. The function would return a filename and allow you modify it. Simpler API/design.

lcxfs1991: I also have to pass filename format  like `[name].css`. Is is possible to do that via a `function`?

README.md
lcxfs1991: what does that mean?

bebraw: It's a typo.

README.md
bebraw: You can return a pattern and it will work like before. You can do `filename: file => '[name].css'` in this design too. And access to the `file` parameter gives you flexibility for custom logic.

lcxfs1991: If you add `js/a` as the entry key, webpack will replace [name] with `css/js/a` and it leaves no room for developers to do the following action:

```js
filename.replace('js/css', 'css');
```

lcxfs1991: you can try my test case:

test/cases/multiple-entries-filename/webpack.config.js

bebraw: I see. It's that `compilation.getPath` which is the problem. Basically `modify` triggers **after** that.

The design could be inverted like this:

```javascript
filename: (getPath, filename) => getPath(filename)
```

That `getPath` would run `compilation.getPath` with all its parameters and provide the hook that would give you maximum amount of control. Now you could manipulate `filename` before and after placeholders get applied while keeping the API surface minimal.

lcxfs1991: Expose `getPath` is cool too.

let me retune it.

lcxfs1991: look good?

extra

bebraw: Yeah, I'll have a closer look in a bit. 👍

bebraw:
This looks good to me.

Please resolve that conflict (had to merge a readme fix). I'll merge this for 2.1 (first release after final 2.0).

bebraw:
@d3viant0ne Do you want to take a quick peek? Looks good to me for a quick 2.1.

bebraw:
Published as `2.1.0`. Sorry for the delay. Busy weeks.

PR #431, opened 2017-02-24
Fix "Multiple Instances" example

daltones:
Both source types `.css` and `.less` will result in pure CSS content after `css-loader`. So it's expected that the extension for the extracted file will always be `.css`.


bebraw:
Yeah, makes sense. Thanks. 👍

PR #443, opened 2017-03-02
Added two tests to extract

chentsulin:
Added testcase for :

```js
ExtractTextPlugin.extract({
  use: ["css-loader", "postcss-loader", "sass-loader"]
});
```

and

```js
ExtractTextPlugin.extract({
  use: [
    { loader: "css-loader", options: { modules: true } },
    "postcss-loader"
  ]
});
```

to make sure them work well.

### DIFF
--- a/test/extract.test.js
+++ b/test/extract.test.js
@@ -72,12 +72,36 @@ describe("ExtractTextPlugin.extract()", function() {
 			]);
 		});
 
+		it("accepts an array of loader names in loader object", function() {
+			ExtractTextPlugin.extract({
+				use: ["css-loader", "postcss-loader", "sass-loader"]
+			}).should.deepEqual([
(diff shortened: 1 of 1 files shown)


chentsulin:
@TheLarkInn I can't get function work with `use` because `use` only supports `"type": ["string", "array", "object"]` in `schema/loader-schema.js`:

https://github.com/webpack-contrib/extract-text-webpack-plugin/blob/master/schema/loader-schema.js#L12-Lundefined

so I got a `Error: data.use should be string,array,object`.

Should I try to change it for pass the tests like this one?

```js
it("accepts an `use` function which return a loader string", function () {
    ExtractTextPlugin.extract({
        use: function() { return "css-loader"; }
    }).should.deepEqual([
       { loader: loaderPath, options: { omit: 0, remove:true } },
       { loader: "css-loader" }
    ]);
});
```

PR #452, opened 2017-03-06
fix: remove module from chunk

sresant:
Remove the module from the chunk when merging non initial chunks...this line seemed to have been removed at some point during some refactoring(?). Removing this line meant that the css files included modules that should not have been included in the bundle making the file sizes much larger then they should be. I noticed this issue on webpack 2

EDIT: Found the commit that removed this line https://github.com/webpack-contrib/extract-text-webpack-plugin/commit/bba0d3c9ae0bd8d11af25196b43952caa8ebcf92#diff-168726dbe96b3ce427e7fedce31bb0bc ...anyone know what bug it fixed


hccampos:
I can confirm that this fixes the issues I've been having with running out of memory with webpack.

tomasztomys:
I have the same issue. Please merge and release

sokra:
The bug is tested by the test case you touched in your PR.

b.txt should include `a`.

sresant:
@sokra Yea I had to remove the b.txt file there. It seemed to me that it made sense that only a.txt was produced since b.txt === a.txt so they should have been merged?

michael-ciniawsky:
@sresant So with the test case change it would incorrect behaviour and this isn't working?

sresant:
@michael-ciniawsky The test fails with the change set. I had to remove b.txt for that test to pass.

michael-ciniawsky:
@sresant But does it fail 'correctly', because the test case is needed and it shouldn't be removed, or because it was wrong aswell ? 😛 I'm confused

sresant:
@michael-ciniawsky Yea that is what I'm not sure about...not much documentation there but from the title of the test dir I had thought that the expected result was incorrect since the two expected files should have been merged into a single one?

tomasztomys:
please merge it, I need this changes

michael-ciniawsky:
@tomasztomys I haven't had time to test this, but you can help me by giving some insight why you need it and what issue you have atm :)

d3viant0ne:
@tomasztomys - This will be merged when it is ready to be merged. 

Currently there is a discussion about a test change you can see above that is blocking this pull request from landing and will continue to until it's resolved.

zigomir:
I'm toying around with Vue that uses this plugin to extract all CSS code into a file using `extractCSS: true` property. What I figured out is that whenever we bundle all JS in one file all goes well, but whenever I split to have vendor bundle separated from other entries vendor file will include modules that will inject styles in runtime, which aren't needed in this case. I tried applying changes from this branch to my locally installed `node_modules/extract-text-webpack-plugin` but it didn't fix the issue. Should I look somewhere else? thanks!

test/cases/merging-multiple-entries-shared/expected/b.txt
```diff
@@ -1 +0,0 @@
-a
```
d3viant0ne: As Tobias mentioned, this test case is part of the validations for the code change in the pull request. `a` in this case should exist in `b.txt` as it is part of the use case for merging multiple shared entries.

sresant:
I probably wont have much time to look at this for the next couple weeks if someone wants to go ahead and dig deeper into this feel free!

michael-ciniawsky:
https://github.com/webpack-contrib/extract-text-webpack-plugin/pull/508/files maybe resolved this (not 💯)

Issue #479, opened 2017-04-12
Reduce package size


evilebottnawi:
Package contains unnecessary directories and files - `.github`, `logo.svg` and others. I can do PR if your accept this issue.

michael-ciniawsky:
@evilebottnawi go ahead 😛

michael-ciniawsky:
Closed by #482

PR #482, opened 2017-04-13
Chore: reduce package size.

evilebottnawi:
Ref: https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/479

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,0 @@
-example/
-test/
-.gitattributes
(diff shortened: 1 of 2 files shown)


michael-ciniawsky:
@evilebottnawi via `package.json`.files please :)

evilebottnawi:
@michael-ciniawsky I deleted `.npmignore` because we don't need this now, also we can add `main` field in this  PR

michael-ciniawsky:
In regards of `package.json`.main please ensure it not causes any trouble on the edges (shouldn't, but who knows 😛)

evilebottnawi:
@michael-ciniawsky tested, all works file :smile:

michael-ciniawsky:
@evilebottnawi Thx

Issue #477, opened 2017-04-09
refactor: apply webpack-defaults


d3viant0ne:
Addition of `webpack-defaults` & associated refactoring as a part of the next `Major` release

Issue exists for status tracking across the organization. 

>Please do not close

d3viant0ne:
Closed with 3e78452

PR #483, opened 2017-04-13
refactor: pass a unique compiler name to get child compilation (webpack 3)

### LINKED ISSUE #389: Extract Text Plugin does not work in child compilations
From https://github.com/jantimon/html-webpack-plugin/issues/579

The root problem is that the extract text plugin uses kind of a hack to communicate between loader and plugin. (See here https://github.com/webpack-contrib/extract-text-webpack-plugin/blob/master/index.js#L216-L227 and here https://github.com/webpack-contrib/extract-text-webpack-plugin/blob/master/loader.js#L26)
This hack works only in the main compilation however the html-webpack-plugin is using a child compilation.

This might add a feature request for the webpack core (easier communication between loaders and plugins)

sokra:
This is recommended after https://github.com/webpack/webpack/pull/4704

It's not good to merge before this PR because it would create absolute paths in the records.

Closes #389

### DIFF
--- a/loader.js
+++ b/loader.js
@@ -49,20 +49,12 @@ module.exports.pitch = function(request) {
 			filename: childFilename,
 			publicPath: publicPath
 		};
-		var childCompiler = this._compilation.createChildCompiler("extract-text-webpack-plugin", outputOptions);
+		var childCompiler = this._compilation.createChildCompiler("extract-text-webpack-plugin " + NS + " " + request, outputOptions);
 		childCompiler.apply(new NodeTemplatePlugin(outputOptions));
 		childCompiler.apply(new LibraryTemplatePlugin(null, "commonjs2"));
 		childCompiler.apply(new NodeTargetPlugin());
 		childCompiler.apply(new SingleEntryPlugin(this.context, "!!" + request));
 		childCompiler.apply(new LimitChunkCountPlugin({ maxChunks: 1 }));
-		var subCache = "subcache " + NS + " " + request; // eslint-disable-line no-path-concat
-		childCompiler.plugin("compilation", function(compilation) {
-			if(compilation.cache) {
-				if(!compilation.cache[subCache])
-					compilation.cache[subCache] = {};
-				compilation.cache = compilation.cache[subCache];
-			}
-		});
 		// We set loaderContext[NS] = false to indicate we already in
 		// a child compiler so we don't spawn another child compilers from there.
 		childCompiler.plugin("this-compilation", function(compilation) {


d3viant0ne:
Note to @webpack-contrib/org-maintainers - Should be done in conjunction with #477 as a single `semver: Major` to coincide with the `webpack@3.0.0` release.

sokra:
@bstuff It's not automatically fixed. This webpack PR only allow to use the `compiler` matching in `rules`:

``` js
rules: [
  { test: /\.css$/, oneOf: [
    { compiler: "html-webpack-plugin", loader: "null-loader" },
    { use: ETP.extract(...) }
  ] }
]
```

sokra:
btw. html-webpack-plugin can do a similar change for webpack 3. cc @jantimon

marmelin:
Hi guys, 
I'm using webpack@3.8.1 and still meet this problem. 
`Module build failed: Error: "extract-text-webpack-plugin" loader is used without the corresponding plugin`

The rules-section looks like this: 
```css
 // CSS LOADER
      , {
          test: /\.css$/
        , use: ['css-to-string-loader'].concat(
              ExtractTextPlugin.extract({
                  fallback: 'style-loader'
                , use: ['css-loader']
              })
            )
        }
```
How ca I implement the above "oneOf" workaround into my loader config?
Is this still necessary?

Thanks.

michael-ciniawsky:
@marmelin remove `css-to-string-loader`

marmelin:
Thanks, 
now without `css-to-string-loader` the configuration looks like this:
**In config file :**
```
...
, { test: /\.css$/
   , use:  ExtractTextPlugin.extract({
           fallback: 'style-loader'
         , use: ['css-loader']
       })
 } ...
```
**as merged object in Config-object output :**
```
...
{ test: /\.css$/,
  use:
   [ { loader: '/home/python1/Dokumente/development/git/v2-web-client/node_modules/extract-text-webpack-plugin/dist/loader.js',
       options: [Object] },
     { loader: 'style-loader' },
     { loader: 'css-loader' } ] }
```
Error is still the same. 

Greetings.

kimown:
@marmelin I met the same problem, do you have any solution finally, thanks.

marmelin:
Hi @kimown , 
I solved this error along with some others and so I forgot to mention it here :) . 

See my repo [here:marmelin/ngtools-webpack-error-message-undefined](https://github.com/marmelin/ngtools-webpack-error-message-undefined), where I solved the error and explain it.

It boils down to this: 

> You can't use extract-text-webpack-plugin for components CSS templates, because it does not support / or do not function with "Child compilations". However you can with global css. Use two CSS rules in webpack config: 1) with "css-loader" and second with extract-text-webpack-plugin and include statement, for your global CSS files.

```
  // COMPONENT CSS LOADER
      , { test: /\.css$/
        , use: ['css-to-string-loader'].concat(['css-loader'])
        }
      // GLOBAL CSS LOADER
      , { test: /\.css$/
         , include: helpers.getPath('src/client/style')
         , use: ['css-to-string-loader'].concat(
            ExtractTextPlugin.extract({
                fallback: 'style-loader'
              , use: ['css-loader']
            })
          )
        }
```
This is what I did and it solved the error!
Good luck :)

PR #500, opened 2017-05-10
Fix table layout

langri-sha:
A misplaced `|` was breaking the row.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ new ExtractTextPlugin(options: filename | object)
 |Name|Type|Description|
 |:--:|:--:|:----------|
 |**`id`**|`{String}`|Unique ident for this plugin instance. (For advanced usage only, by default automatically generated)|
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,7 +68,7 @@ new ExtractTextPlugin(options: filename | object)
 |Name|Type|Description|
 |:--:|:--:|:----------|
 |**`id`**|`{String}`|Unique ident for this plugin instance. (For advanced usage only, by default automatically generated)|
-|**`filename`**|`{String\|Function}`|Name of the result file. May contain `[name]`, `[id]` and `[contenthash]`|
+|**`filename`**|`{String/Function}`|Name of the result file. May contain `[name]`, `[id]` and `[contenthash]`|
```
michael-ciniawsky: `{String\|Function}` is the correct behaviour, why do you want to change it ? 😛

Issue #459, opened 2017-03-15
Using extract-text & commons-chunk together, the css output for entry may or may not exist


hanxinwei:
depending on if the entry needs at lease one more css module.
I guess this is by design.
However this is kind of inconvenient, for the server have to check if the entry css exists and decide whether to append it.

Compare the entry js to css
js: If the entry js does nothing ( even if totally empty ), the output entry file still be generated.
css: if the entry css needs no more injection, there is no output file.

Is that possible to provide a option to generate empty entry css file?

Example repo
https://github.com/hanxinwei/extract-text-commons-chunk

Source

/src
  /css
     style.css
     style2.css
  /scripts
    /apps
      page.js
      page2.js
   /services
      service.js
      service2.js

Page.js
import service from '../services/service';
import service2 from '../services/service2';
import '../../css/style.css';
import '../../css/style2.css';

page2.js
import '../../css/style.css';

Config

module.exports = {
    entry: {
        page: './src/scripts/apps/page.js',
        page2: './src/scripts/apps/page2.js'
    },
    output: {
        path: path.resolve(__dirname, './build/'),
        filename: 'scripts/[name].js'
    },
    module:
    {
        rules: [
            {
                test: /\.css$/,
                use: ExtractTextPlugin.extract('css-loader')
            }
        ]
    },
    plugins: [
        new webpack.optimize.CommonsChunkPlugin({
            name: 'common'
        }),
        new ExtractTextPlugin('css/[name].css')
    ]
}

Output
/build
   /scripts
     common.js
     page.js
     page2.js
   /css
     common.css
     page.css

There is no page2.css generated.

aciccarello:
I am also running into an issue using both the ExtractTextPlugin and CommonsChunkPlugin as well. I noticed that css imported by angular-ui-bootstrap, which is included in a vendor bundle, was not being displayed.

After doing some debugging it seems that any css/scss in my project which is imported in the commons chunk will not appear in the project (either in the extracted css file or in the common module). CSS that is not imported by the vendor bundle does appear in the extracted css file. The problem goes away when the ExtractTextPlugin is disabled.

### Entries
```javascript
entry: {
  vendor: './vendor.js',
  main: './main.js'
}
```

### Rules
```javascript
  {
    test: /\.scss$/,
    loader: ExtractTextPlugin.extract({
      fallback: 'style-loader',
      use: ['css-loader', 'sass-loader']
    })
  },

  {
[...]

Issue #120, opened 2015-10-26
ExtractTextPlugin extracts from asynchronous split point chunks


chrispesto:
I have a webpack.config.js like:

```
module.exports = {
  entry: {
    index: ['./app/routes/ClientRouter', './app/pages/Index'],
    create:  ['./app/routes/ClientRouter', './app/pages/Create'],
    play:  ['./app/routes/ClientRouter', './app/pages/Play'],
    vendor: ['react', 'react-dom', 'flux']
  },
  output: {
    path: __dirname + '/public/build',
    publicPath: '/build/',
    filename: '[name].js',
    chunkFileName: '[id].chunk.js'
  },
  resolve: {
    alias: alias
  },
  module: {
    noParse: noParse,
    loaders: [
      { test: /.js$/, loader: 'babel-loader', exclude: /node_modules/ },
      { test: /.css$/, loader: ExtractTextPlugin.extract('style-loader', 'css-loader'), exclude: /node_modules/ }
    ]
  },
  plugins: [
    new webpack.optimize.CommonsChunkPlugin({ name: 'vendor', filename: 'vendor.js', minChunks: Infinity }),
    new ExtractTextPlugin('[name].css', {
      allChunks: false
    })
  ]
};
```

And a client router like:

```
function route() {
  var location = window.location.pathname,
      appContainer = document.getElementById('app-container'),
      routeMatches;

  if(!location || location == '/') {
    require.ensure(['../pages/Index'], function() {
      var Page = require('../pages/Index');
      ReactDOM.render(<Page />, appContainer);
    });
  } else if(routeMatches = location.match(/^\/create$/)) {
    require.ensure(['../pages/Create'], function() {
      var Page = require('../pages/Create');
      ReactDOM.render(<Page />, appContainer);
    });
  } else if(routeMatches = location.match(/^\/play\/([01]+)$/)) {
    require.ensure(['../pages/Play'], function() {
      var Page = require('../pages/Play');
      ReactDOM.render(<Page mazeData={routeMatches[1]}/>, appContainer);
    });
  }
}
```

The ExtractTextPlugin creates index.js/index.css, create.js/create.css, etc, with the extracted CSS for each containing only what is directly required from each of those entry points. It also generates 1.1.js, 2.2.js, and 3.3.js for the split points to load pages asynchronously, but the CSS is removed from them as well. I.e. (from 1.1.js):

```
/***/ 5:
/***/ function(module, exports) {

[...]

PR #507, opened 2017-05-18
fix not extracting CSS from common async chunks

numical:
**The problem**
1. Create an entry point requiring *two* asynchronous chunks;
2. Name these asynchronous chunks;
3. Configure CommonsChunkPlugin to generate a common chunk explicitly for these two asynchronous chunks;
4. If the resultant commons chunk has embedded CSS, ExtractTextPlugin does not extract this.

See the added test case in this PR for the scenario.

This may be related to issues [120](https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/120) and [459](https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/459).

**The cause**
The generated async chunk has no parents and no entry point.

**The fix**
Replace all uses of `chunk.isInitial()` with `chunk.isInitial() || chunk.parents.length === 0`.

(Sorry for the mutliple commits - intially committed as the wrong user).


evilebottnawi:
@numical be great add tests

evilebottnawi:
@numical please accept `licence/cla`, thanks

numical:
I have.
Unfortunately my initial commit was under the wrong git user name.
Hence the subsequent commits under the correct 'numical' account.
Will it be easier if I cancel this PR and generate a cleaner one?

evilebottnawi:
@numical yep, it would be perfect

numical:
Will do

Issue #461, opened 2017-03-20
TypeError: chunk.isInitial is not a function


jony89:
versions :
    
     "webpack": "^2.2.1"
    "extract-text-webpack-plugin": "^2.1.0",

error : 

> \node_modules\extract-text-webpack-plugin\index.js:267
> 				var shouldExtract = !!(options.allChunks || chunk.isInitial());
> 				                                                  ^
> 
> TypeError: chunk.isInitial is not a function
>     at \node_modules\extract-text-webpack-plugin\index.js:267:55
>     at \node_modules\extract-text-webpack-plugin\node_modules\async\dist\async.js:3025:16
>     at eachOfArrayLike (\node_modules\extract-text-webpack-plugin\node_modules\async\dist\async.js:940:9)
>     at eachOf (\node_modules\extract-text-webpack-plugin\node_modules\async\dist\async.js:990:5)
>     at Object.eachLimit (\node_modules\extract-text-webpack-plugin\node_modules\async\dist\async.js:3089:3)
>     at ExtractTextPlugin.<anonymous> (\node_modules\extract-text-webpack-plugin\index.js:265:10)
>     at Compilation.applyPluginsAsync (\node_modules\gulp-webpack\node_modules\tapable\lib\Tapable.js:71:13)
>     at Compilation.seal (\node_modules\gulp-webpack\node_modules\webpack\lib\Compilation.js:525:7)
>     at Compiler.<anonymous> (\node_modules\gulp-webpack\node_modules\webpack\lib\Compiler.js:397:15)
>     at \node_modules\gulp-webpack\node_modules\tapable\lib\Tapable.js:103:11
>     at Compilation.<anonymous> (\node_modules\gulp-webpack\node_modules\webpack\lib\Compilation.js:445:10)
>     at \node_modules\gulp-webpack\node_modules\webpack\lib\Compilation.js:417:12
>     at \node_modules\gulp-webpack\node_modules\webpack\lib\Compilation.js:332:10
>     at \node_modules\async\lib\async.js:52:16
>     at done (\node_modules\async\lib\async.js:246:17)
>     at \node_modules\async\lib\async.js:44:16

for some reason I could not attach the ZIP to the message, so here a link to simple standalone project : https://ufile.io/596a0

bebraw:
Here's the project - https://github.com/jony89/simple-webpack-extract-plugin .

I tried this on macOS (Node 6.9.1) and it completed without any problems.

[...]

Issue #449, opened 2017-03-05
TypeError: chunk.isInitial is not a function


daryn-k:
There is this error:
```
var shouldExtract = !!(options.allChunks || chunk.isInitial());
TypeError: chunk.isInitial is not a function

```
I have no idea what is it.

My webpack.config.js file:

```
var ExtractTextPlugin = require('extract-text-webpack-plugin');

module.exports = {
    entry: './app/client.js',
    output: {
        filename: 'bundle.js',
        path: 'public/js'
    },
    module: {
        loaders: [
            {
                test: /\.jsx?$/,
                exclude: /node_modules/,
                loader: 'babel-loader',
                query: {
                    presets: ['react', 'es2015', 'stage-1']
                }
            },
            {
                test: /\.css/,
                loader: ExtractTextPlugin.extract('css')
            }
        ]
    },
    plugins: [
        new ExtractTextPlugin('styles.css')
    ]
};
```

bebraw:
There's not enough information to debug. Which version of webpack/extract-text-webpack-plugin are you using? If you have webpack 1 and are using extract-text-webpack-plugin 2, that error could happen. In that case you'll have to use extract-text-webpack-plugin 1 or upgrade webpack to 2.

daryn-k:
I use last versions of 'extract-text-webpack-plugin' and webpack (2.2.1)

bebraw:
Ok, still not enough info. Maybe someone else has a better idea.

daryn-k:
I don't know what else... It's  just about a simple functionality - build 'bundle.css' from all css-files from project. What kind of information do you need?

bebraw:
I would need a project to run. `chunk.isInitial()` is something provided by webpack itself. If `chunk` is missing, there is something fatally wrong.

daryn-k:
Maybe these lines?

```
 at C:\site.dev\site-spa\react\node_modules\extract-text-webpack-plugin\index.js:259:55
    at C:\site.dev\site-spa\react\node_modules\extract-text-webpack-plugin\node_modules\async\dist\async.js:3025:16
    at eachOfArrayLike (C:\site.dev\site-spa\react\node_modules\extract-text-webpack-plugin\node_modules\async\dist\async.js:940:9)
    at eachOf (C:\site.dev\site-spa\react\node_modules\extract-text-webpack-plugin\node_modules\async\dist\async.js:990:5)
[...]

Issue #261, opened 2016-10-05
chunk.isInitial is not a function


Awk34:
I'm getting this error when trying to build my app:

```
var shouldExtract = !!(options.allChunks || chunk.isInitial());
                                                  ^

TypeError: chunk.isInitial is not a function
    at <myApp>\node_modules\extract-text-webpack-plugin\index.js:246:55
    ...
```

I'm using `webpack`@`2.1.0-beta.25` & `extract-text-webpack-plugin`@`2.0.0-beta.4`

I can provide more details if they're needed

Awk34:
My fault. I was using `webpack-stream`, which I wasn't aware includes it's own dependency on Webpack@1. Running Webpack from the Node interface fixed this issue.

allahbaksh:
What do you mean running "webpack from node interface". I am having similar problem

Roy9102:
@allahbaksh It means your global webpack (terminal cli) is the webpack@1.xxx, you need to upgrade your webpack,  you can check out your webpack version by 'webpack -v'

Awk34:
@allahbaksh https://webpack.js.org/api/node/

PR #512, opened 2017-05-22
Add undefined check

znck:
Sometimes last entry in `chunks` is `undefined`.

I'm investigating the cause until this fix is a work around.

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,7 +25,7 @@ ExtractTextPlugin.prototype.mergeNonInitialChunks = function(chunk, intoChunk, c
 	if(!intoChunk) {
 		checkedChunks = [];
 		chunk.chunks.forEach(function(c) {
-			if(c.isInitial()) return;
+			if(c === undefined || c.isInitial()) return;
 			this.mergeNonInitialChunks(c, chunk, checkedChunks);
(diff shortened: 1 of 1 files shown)


evilebottnawi:
@znck it should be not `undefined`, seems your have invalidate configuration or bad plugin/loader behavior

evilebottnawi:
@znck be good in future first create issue (with minimal test repo, include config, - what expected and actually happens) and then do PR. Thanks!

znck:
@evilebottnawi There are multiple similar issues - #449 #461 #261 

I know this is a side-effect of some other loader/config.

evilebottnawi:
@znck be right found where problems and fix it

evilebottnawi:
Sorry accidentally closed, could you give minimal reproducible test repo?

znck:
Closing it. Will try to create a repro!

znck:
Cause: Extra entry in `chunks` option of CommonsChunkPlugin.

Example: 
```
CommonsChunkPlugin({
    name: 'foo',
    chunks: [
      'app', 
      'random-chunk-name' // Removing this fixed the issue.
    ],
  }
```

PR #520, opened 2017-05-30
Add ignoreOrder check

keepitreal:
This PR adds the ignoreOrder checks added by #166 in the master branch. I realize the optimum solution for this issue is to migrate to Webpack 2 but since I'd consider this issue fairly debilitating, I think a fix should be present for those who need something in their production environments right away and can't go through the steps to migrate.

### DIFF
--- a/index.js
+++ b/index.js
@@ -289,7 +289,7 @@ ExtractTextPlugin.prototype.apply = function(compiler) {
 			extractedChunks.forEach(function(extractedChunk) {
 				if(extractedChunk.modules.length) {
 					extractedChunk.modules.sort(function(a, b) {
-						if(isInvalidOrder(a, b)) {
+						if(!options.ignoreOrder && isInvalidOrder(a, b)) {
 							compilation.errors.push(new OrderUndefinedError(a.getOriginalModule()));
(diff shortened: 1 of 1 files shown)


evilebottnawi:
@keepitreal Support legacy code (options, functions, checks, and etc) is a bad practice (we use semver to avoid accumulation this legacy stuff). I would advise you to find the time and go to `webpack` v2 or do fork and use this fork for own packages, where your use `webpack` v1. Not here. Thanks!

michael-ciniawsky:
@bebraw @d3viant0ne We can still cut a release for `v1.x`?  The question is if we want to take the eventual risk that this might break something in general, if not then better close the PR :)

d3viant0ne:
imo, given that we are about to release 3.x the time for adding features for Webpack 1.x has come and gone.

TheLarkInn:
Agreed.

d3viant0ne:
And for reference, 1.x is already not supported in the current release version.

screen shot 2017-06-07 at 9 38 53 pm

d3viant0ne:
@keepitreal - If you need a hand getting that upgraded shoot us an @ and we'll get you squared away.

PR #532, opened 2017-06-08
docs: rm `RELEASE.md`

michael-ciniawsky:
I still have the file, but I think it's better placed somewhere else (e.g webpack-contrib/organisation) or the like, since it's an 'internal' matter of affairs :)


d3viant0ne:
Yeah, it was moved into the docs & then removed. Permanent home will be in the contrib sections of webpack.js.org

PR #538, opened 2017-06-09
docs(README): use warning emoji

langri-sha:
Fixes display of `:warning:` on https://webpack.js.org/plugins/extract-text-webpack-plugin/#usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ new ExtractTextPlugin(options: filename | object)
 * `[id]` number of the chunk
 * `[contenthash]` hash of the content of the extracted file
 
-> :warning: `ExtractTextPlugin` generates a file **per entry**, so you must use `[name]`, `[id]` or `[contenthash]` when using multiple entries.
+> ⚠ `ExtractTextPlugin` generates a file **per entry**, so you must use `[name]`, `[id]` or `[contenthash]` when using multiple entries.
 
 #### `#extract`
 


michael-ciniawsky:
warning

@langri-sha Where does `:warning:` currently fail to display correctly ?

michael-ciniawsky:
Feel free to reopen if still regressions and more info provided :)

langri-sha:
Not sure where are you looking at. I'm looking at the document provided in the link.

image

☝️

michael-ciniawsky:
cc @bebraw @skipjack Possible to support Github Emoji on webpack.js.org?

bebraw:
@michael-ciniawsky I think we would have to use an emoji fork of marked or some other solution to get that. I had to solve the same problem yesterday and it was a little hacky.

michael-ciniawsky:
markdown-it && e.g [markdown-it-emoji](https://www.npmjs.com/package/markdown-it-emoji) ?

bebraw:
@michael-ciniawsky That could work.

skipjack:
Does markdown-it support all the other customization we have? Or, even better, does it support similar features to github's parser so we don't have to do as much customization?

michael-ciniawsky:
Depends on what customization are currently required 😛 , ther are many plugins and writing one isn't that hard either, `remark` maybe a valueable drop in replacement in case the plugin ecosystem there is 'brighter'. Both are definitely better then `marked`  imho :)

skipjack:
Ok that's good to know -- @bebraw maybe we should open an issue on [webpack/webpack-defaults](https://github.com/webpack-contrib/webpack-defaults) to start discussing this more? Whatever is decided there could influence a follow-up issue and prs in the docs repo. Maybe move our existing `.markdownlint` over there as a baseline and tweak it further after the discussion?

bebraw:
> Ok that's good to know. @bebraw maybe we should open an issue on webpack/webpack-defaults to start discussing this more? Whatever is decided there could influence a follow-up issue and prs in the docs repo. Maybe copy our existing .markdownlint over there as a baseline and tweak it further after the discussion?

Yeah, please open an issue and refer to there from here. It's better to solve anything readme related in one place (the point of defaults).

michael-ciniawsky:
Closing in favour of [#73 (`webpack-defaults`)](https://github.com/webpack-contrib/webpack-defaults/issues/73). Thx

PR #543, opened 2017-06-13
refactor: `Chunk.modules` => `Chunk.mapModules`  (`webpack v3.0.0`)

MirrorBytes:
Refactors to use webpack 3.x `Chunk.mapModules` 

Closes #529 

> Squashed commit message body for merge
```

- refactor: DeprecationWarning: Chunk.modules

BREAKING CHANGE: Updates to `Chunk.mapModules`. This release is not backwards compatible with `Webpack 2.x` due to breaking changes in webpack/webpack#4764
```

PR #542, opened 2017-06-12
refactor: Apply webpack-defaults

d3viant0ne:
- [x] Apply `webpack-defaults` tooling updates
- [x] Apply `webpack-defaults` grammar updates
- [x] Apply `webpack-defaults` style updates
- [x] Convert existing tests to Jest from Mocha

The desired effect here is to `update` style & grammar without changing the way etwp works.

Note, this is being merged into #540 making this refactoring `etwp & webpack v3 only`

Closes #477

PR #540, opened 2017-06-11
refactor: Apply webpack-defaults & webpack 3.x support

d3viant0ne:
Working PR for the 3.x release branch

- [x] Pick up child compilation fixes from webpack/webpack#4704
- [x] Fix deprecation warnings in `chunk.modules` from webpack/webpack#4764 #543
- [x] Apply `webpack-defaults` tooling updates #542
- [x] Apply `webpack-defaults` grammar updates #542
- [x] Apply `webpack-defaults` style updates #542
- [x] Convert existing tests to Jest from Mocha #542
- [x] Enable Appveyor service #542 
- [x] Remove the hard `webpack` devDependency #542 
- [x] Remove both `single argument` deprecations from 2.x #542
- [x] Update the `Install` section of readme with last  published 2.x version
- [x] Fix the css generation order issue in #563
- [x] TypeError: [object Object] is not a function #565

**Beta & RC builds should be published from this branch** to the `@beta` npm `dist-tag`

> Squashed commit message body for merge and `3.0.0` release
```

- refactor: Pass a unique compiler name to get child compilation [483](https://github.com/webpack-contrib/extract-text-webpack-plugin/pull/483)
- refactor: Apply webpack-defaults [542](https://github.com/webpack-contrib/extract-text-webpack-plugin/pull/542)

BREAKING CHANGE: Enforces `engines` of `"node": ">=4.3.0 < 5.0.0 || >= 5.10`

- refactor: DeprecationWarning: Chunk.modules [543](https://github.com/webpack-contrib/extract-text-webpack-plugin/pull/543)

BREAKING CHANGE: Updates to `Chunk.mapModules`. This release is not backwards compatible with `Webpack 2.x` due to breaking changes in webpack/webpack#4764

- fix: css generation order issue see: webpack/webpack#5225

BREAKING CHANGE: Enforces `peerDependencies` of `"webpack": "^3.1.0"`. 

```

Closes #529
Closes #548

Issue #485, opened 2017-04-17
 Cannot find module 'webpack/lib/Chunk'


manuelro:
I'm not even using the plugin and I get the following error when I run `webpack` from my terminal:
```
module.js:471
    throw err;
    ^

Error: Cannot find module 'webpack/lib/Chunk'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> ([absolute path to project folder]\node_modules\extract-text-webpack-plugin\index.js:9:13)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
```
The plugin is loading a module called `Chunk` (see line 9 on the index file):
`var Chunk = require("webpack/lib/Chunk"); // Line 9 @ extract-text-webpack-plugin/index.js`

But there is not such folder in Webpack module:
![image](https://cloud.githubusercontent.com/assets/4284074/25100705/5bc710fc-236e-11e7-81e3-3a29b6576063.png)

Any ideas of why this is happening?

Webpack version: `v2.3.3`
Plugin version: `2.1.0`
NPM version: `3.10.9`
Node version: `v6.9.2`

manuelro:
UPDATE: Installing a previous version of Webpack solved the issue.

`npm install webpack@2.2.0 --save-dev`

radix:
why did you close this? is there a way to use this plugin with recent versions of webpack?

DanielTate:
This is still an issue on Webpack 2.6

manuelro:
I just reopened this because there seem to be issues with recent Webpack versions as well. As I stated it before, I fixed this by using a previous version (`npm install webpack@2.2.0 --save-dev`). But clearly this was not an optimal solution.

b4dnewz:
``` bash
Error: Cannot find module 'webpack/lib/Chunk'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    ...
```

using __webpack version 3.4.1__ with the snippet from the readme.

```javascript
const path = require('path');
[...]

PR #546, opened 2017-06-17
fix: split async CSS correctly

BowlingX:
This is a work-in-progress pull-request to address issue #120.
I added a test case to demonstrate the issue, use-case and my expectations.


BowlingX:
I adjusted the extraction logic, the result is now as expected, however another test (`splitted-multiple-entries`) fails now. I'm not sure if I'm heading in the right direction or if we may should make this configurable.

BowlingX:
ok, ignore the last commit, does not make any sense and is not working at all.

BowlingX:
I think I finally found the correct way to do it. It was a little hard to get my head around it without much docs about the webpack internals, but I think it's a proper solution.

I will now actually keep the fallback loader in place for the async bundles and only rebuild it without replacing it in the async chunks. That means all css is correctly async loaded and only the css for the initial chunk is written down (if `allChunks: false`).

The initial chunks is freed from all dependencies and replaced with the generated module (that may contain the css-module informations).

michael-ciniawsky:
Fixes #120

michael-ciniawsky:
@BowlingX WoW..😧 😛 . I'm not 💯 skilled to answer if this is suitable or not, but it looks very good

evilebottnawi:
@BowlingX please squash all commits in one for easy review, thanks

michael-ciniawsky:
An additional 'problem' is the webpack 3 release, which requires a breaking change here #540 #543 (#542), so we need to fix this first (as quickly as possible). So maybe you need to rework a few parts because of this in any case 🙏

d3viant0ne:
We can always back port it to the 2.x version range after the fact.

BowlingX:
I have a question about the code-style, I used arrow functions but did not use const for example where I could use it. What is the way to go? Stick to ES5 or is it ok to use ES6 features? What node version is the target?

d3viant0ne:
@BowlingX - See #542 which applies `webpack-defaults` which amongst other things, enforces the contrib code style.

See the babelrc & engines objects for the webpack node range as well as how everything transpiles down via babel.

d3viant0ne:
And also for reference, all of this depends entirely on @sokra's approval to move forward. He has a master plan for CSS in webpack which will eventually eliminate etwp.

Regardless, you will need to target the branch I mentioned above as opposed to master. All of the `etwp` beta / rc publishing will be done from this branch with the eventual `3.0.0` being published after #540 (feature/webpack3) lands in master.

src/index.js
```diff
@@ -230,7 +230,9 @@ class ExtractTextPlugin {
               chunk.removeModule(data.moduleToRemove);
               const deps = data.moduleToRemove.dependencies;
               deps.forEach((d) => {
-                chunk.removeModule(d.module);
+                if (d.module && d.module.loaders.length > 0) {
+                  chunk.removeModule(d.module);
+                }
```
faceyspacey: I've been thinking about the best way to handle CSS in dynamic chunks. What i've come up with is this: **`import()` needs to request both the js + the css.** 

I.e. 2 requests, rather than what I was doing before with `extract-css-chunks-webpack-plugin` where I made 2 js files (one with style-loader and another called `no_css.js` without style-loader). That adds considerably to the compilation time.

As far as HMR, this simplifies implementation--it's just one way to support.

The most important thing however is this: ***having a different setup in production than you have in development is a bad idea*** (i.e. where in development css is served in js via style-loader, and during production via external stylesheets). Having the same setup (external stylesheets) means going to production has less unforeseen bottlenecks. 

*If it works in development, it works in production.*

Developers new to Webpack expend a lot of wasted energy figuring this stuff out. I.e. fumbling around with file hashes, only embedding a stylesheet in production (whose file hash can be hard to find), etc. It's a frustrating stage in your project, especially when you thought you were done.

Since HMR works on stylesheets, it makes you think why was CSS ever injected from javascript? To save novice developers from pasting a `<link />` tag in addition  to `main.js`? Perhaps it's just legacy because that's how style loader was built first, or EWTP wasn't around in the beginning.

I think we gotta go back to square one. Maybe this is part of @sokra 's "masterplan." 

In short, it's a babel plugin that changes `import()` to `Promise.all([import(), fetchCss()])`. 

And now this plugin is easier than ever. It insures there is a matching CSS file per chunk, no matter whether it's named or dynamic. And it uses a very straightforward mechanism to reload CSS files corresponding to each module:

```
if (module.hot) {
	module.hot.accept();
	if (module.hot.data) {
		require(${loaderUtils.stringifyRequest(this, path.join(__dirname, "hotModuleReplacement.js"))})("${publicPath}", "%%extracted-file%%");
	}
}`;
```

Also, shouldn't compiling without ever moving css into js chunks be faster? Maybe this is part of the masterplan and needs to happen in core somewhere, but if CSS never makes its way into javascript modules/chunks, and is put off to the side early on, then we're talking not just easier for developers, but faster and maybe easier in implementation.

faceyspacey:
Also if its not clear, basically the idea of having style-loader for async chunks but not for your initial entries is incorrect. Stylesheets should be cacheable separately, as they are less likely to change. Or if it's the reverse and the js changes, no matter what more bytes are cached. Also, on initial load if u have served/embedded additional chunks and rendered those chunks correctly with SSR, the css needs to be a stylesheet to avoid flashes of unstyled content for the portions of the page the chunks corresponds to while the js loads. CSS across all chunks needs to be in its own sheet to do this at the highest level. CSS belongs in sheets. We are going against the grain by injecting it from js.

Next: if we have separate sheets but import() does not load them in parallel and resolve them simultaneously when they both complete, the js chunk requests the stylesheets and as a result they will arrive after, leaving a moment of unstyled content once again. 

Small related issue: multiple modules in a js chunk will request the same CSS file (though that can be solved easily enough).

So all that's to say stylesheets must come first always. If the js sequentially loads the CSS, the js can't render on the page until the CSS is there. Loading spinners must continue to soon. This is why the only solution I can see is a customized import() that loads both. Sequential--if done right--could be a stop gap solution if loading both in parallel is far off. But it shouldn't be and we should bite the bullet once and for all. The babel plugin concept can be implemented immediately, so long as we have multiple stylesheets with no style loader.

@BowlingX can u bring back as an option the async stylesheets like we had in extract CSS chunks?

So this is why importing

faceyspacey:
One final thing. I realized what the discrepancy is and why I had a hard time understanding you when we talked on the extract-css-chunks-webpack-plugin Issue:

We are trying to solve 2 different use cases. You either have a *single page app* without SSR, or you have SSR but you aren't fully rendering your async chunks synchronously on the server. I.e. you're rendering them with a spinner so checksums match on the client, and then the client makes an additional request to load the chunk. 

My use case is a complete SSR solution where the chunks are synchronously rendered on the server via [React Universal Component](https://github.com/faceyspacey/react-universal-component) (similar to React Loadable). 

In your case you don't suffer from flashes of unstyled content ("FOUC")--because the HTML string initially embedded in the page *does NOT have the content.* 

Ultimately we need to support both cases to truly get this right. Simultaneous Code-Splitting + SSR is about to get a lot more popular as people starting using [these techniques](https://medium.com/@faceyspacey/code-cracked-for-code-splitting-ssr-in-reactlandia-react-loadable-webpack-flush-chunks-and-1a6b0112a8b8) or making their own. 

So I realize the way we have been developing up until now requires your use-case more, but in my opinion, React world can't go on any longer without true and complete SSR. And it's just not gonna happen now that it's possible and generally available. We need to prepare for this future. We need to make this future. CSS plays a key role in making this all happen. It's a missed opportunity if EWTP and the eventual "master plan" don't make this a first class priority. 

So if you could make dynamic css chunks an option, I'd greatly appreciate it. You're far better than I am working with Webpack code, judging from how you whipped what I did into shape. If you can do that, I can work on the rest. In fact, next week I have something coming out called `Universal.render()` that goes even farther than React Universal Component, and really simplifies universal rendering in a way that doesn't require a framework (ahem, Next.js). I think people are gonna be really excited to see it. It brings universal rendering to not just components, but data fetching like [async-reactor](https://github.com/xtuc/async-reactor) does without SSR.

BowlingX:
> We are trying to solve 2 different use cases. You either have a single page app without SSR, or you have SSR but you aren't fully rendering your async chunks synchronously on the server. I.e. you're rendering them with a spinner so checksums match on the client, and then the client makes an additional request to load the chunk.

I actually have one entrypoint for each page and render them individually with PHP on the server, so nothing async and no checksum match problems. But I understand your usercase better now :)., indeed they are different.

This pull-requests target's only the specific problem mentioned in #120 . I think you are right that we need something that allows universal extraction in an async manner and hot-reload but I'm not sure if this is out of the scope of this plugin right now. And if css becomes a first citizen of webpack, we might be able to archive what you outlined.

> Small related issue: multiple modules in a js chunk will request the same CSS file (though that can be solved easily enough).

I actually see a bigger issue here, because the css file might be big and needs to be fully recompiled. Right now only the css part that has been changed is processed by the style-loader (at least I think so). So hot-reloading the full css file might be not the best way.

> The most important thing however is this: having a different setup in production than you have in development is a bad idea (i.e. where in development css is served in js via style-loader, and during production via external stylesheets). Having the same setup (external stylesheets) means going to production has less unforeseen bottlenecks.

This might get challenging because we usually want best performance on production and most convenient and fast reload while developing, so right now the only option is using `style-loader` for css-modules.

I will think about more options.

faceyspacey:
Per module style compilation is less of an issue now that we are building heavily split sites. Basically each chunk may only contain 4 CSS modules.

The problem essentially hits 2 birds with one stone and they all solve each other.

If u aren't doing much splitting, configure different options :)

Any clues on how to do the async splitting in the latest v3 codebase? 

What about a faster way to inject the CSS file name into dependent modules?

It was all done before in ECCWP. At the end of the day, this is just ECCWP without the js file that has CSS injection, but it would nice to have a version of it based on what u have done.

d3viant0ne:
@evilebottnawi - For reference, if it's a PR where Tobias needs to look at something don't squash it.

And in general, we squash when we merge. Leaving the individual commits while the PR is open makes a change set easier to follow from inception to completion.

michael-ciniawsky:
@BowlingX You are still interested in this I hope :) ? Please wait until #540 landed on master (hopefully after the 'weekend' `beta`), in case no new issues ar found in the current beta,  we can then finally focus on the matter here 😛

BowlingX:
Sure :) I keep on solving the conflicts with the webpack-3 branch. But should I target  `master` after the weekend then?

michael-ciniawsky:
👍 It's not 💯  set yet, when #540 lands on `master` (planed after the weekend, if there is no critical bug found), so keep the `feature/webpack3` branch for know please :).

faceyspacey:
For what it's worth I implemented the dual import via a babel plugin as I described above:

https://medium.com/faceyspacey/webpacks-import-will-soon-fetch-js-css-here-s-how-you-do-it-today-4eb5b4929852

Not sure if EWTP is still against HMR and dynamic splitting. It seems webpack core will get all this as the article from a month ago from sokra describes (funny, i just stumbled on that after I built babel-plugin-dual-import).

Anyway it seems CSS has a good path forward in Webpack world. I'm not sure what role I can play. @BowlingX mastered the EWTP codebase far better than I did. Let me know if HMR + dynamic splitting is of interest round these parts.

nikit-cpp:
Hello, will this PR merged to next release 3.0.1 ?

BowlingX:
I merged all changes with the master branch.

nikit-cpp:
I tried to use it, 
```
    "extract-text-webpack-plugin": "git://github.com/BowlingX/extract-text-webpack-plugin.git#143760a4b60c1c4e74d8211e3d1f687e0a56b4cd",

```
but 
```
Error: Cannot find module 'extract-text-webpack-plugin'
    at Function.Module._resolveFilename (module.js:485:15)
    at Function.Module._load (module.js:437:25)
    at Module.require (module.js:513:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/home/nik/ideaWorkspace/checklist/frontend/webpack.config.js:14:27)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
    at Module.require (module.js:513:17)
    at require (internal/module.js:11:18)
    at requireConfig (/home/nik/ideaWorkspace/checklist/frontend/node_modules/webpack/bin/convert-argv.js:97:18)
    at /home/nik/ideaWorkspace/checklist/frontend/node_modules/webpack/bin/convert-argv.js:104:17
    at Array.forEach (native)
    at module.exports (/home/nik/ideaWorkspace/checklist/frontend/node_modules/webpack/bin/convert-argv.js:102:15)
```

it seems same as in https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/485

BowlingX:
So from my side the pull-request is complete, I'm not sure if there is anything else TODO.

@nikit-cpp I think you can't install it directly from the commit, it's build before pushed to the npm repository.

faceyspacey:
Any thoughts on supporting dynamic css chunks, as per extract-css-chunks-webpack-plugin?

I realize the HMR may be something that is still not desired, but it's mind-boggling that css chunks from dynamic imports isn't desired here (let alone not the prime focus), especially since it's part of the eventual master plan.

I can apply it if that's what everyone agrees.

nikit-cpp:
I have some dummy trouble with second and followed consequently builds, 
first buid is ok, but consequently are produces different (as it appears incorrect js) so it fail page rendering vith vuejs (first build is ok).

I spent several days with trying to solve it but it's too hard.

This is home project with java8, redis(without password on defaul port), maven(wrapper, you actually need only JDK and you shouldn't install maven manually).
Should I to realize nodejs express-like backend for serve static? Or you can run java and install redis?

Steps to reproduce:

git clone git@github.com:nikit-cpp/good-parts.git
git checkout extract-twxt-plugin-issues
./mvnw -DskipTests clean package
# it important don't make cd because there is used relative paths for static
start spring boot server
java -jar frontend/target/frontend-1.0-SNAPSHOT-exec.jar

now build and compile frontend(js, css, ...). It uses 143760a, I built this commit and pused to my repo.
cd frontend
npm i
webpack

Now you can see first correct build http://127.0.0.1:8080/post/edit . This is dynamically loaded module for heavy quill editor.
but if you'll can any css (for example style atg in frontend/src/main/frontend/components/PostList.vue)
juat add newline for example, webpack will rebuild it and you'll see error in console 


But if you restart webpack (CTRL+C) and start it again -- this will works.

BowlingX:
Mhh, I once had that issue, but I solved it a while ago, maybe something
went wrong with the merge. I will have a look tomorrow. Thank you.

Nikita Konev <notifications@github.com> schrieb am Di., 25. Juli 2017,
22:08:

BowlingX:
@nikit-cpp I'm not sure if this is a problem with this plugin actually or if the `vue-style-loader` might need to be adjusted to the new version...

BowlingX:
Seems like vue is handling async css loading on it's own. I think this get's in the way how extrac-text-webpack-plugin and my changes are doing this right now. So I would not expect that to be working together.

nikit-cpp:
@BowlingX 
Thank you for response

I also found that I can fix it by disable webpack's cache (introduced in 3.0.0)

Unfortunately there isn't activity in https://github.com/vuejs/vue-style-loader, ok, I'll try to create issue there (vue-style-loader repo).

BowlingX:
What is the status for this pull-request? Is there any more work needed?

BowlingX:
I was able to reproduce the issue that @nikit-cpp had. Something is not right here, I will investigate

nikit-cpp:
@BowlingX I can check if need

michael-ciniawsky:
The approach looks fine so far (from what i can tell :)) but as mentioned before I can't fully approve the changes. @sokra needs to review this PR to ensure this doesn't break the plugin

src/index.js
```diff
@@ -26,7 +27,7 @@ class ExtractTextPlugin {
     if (isString(options)) {
       options = { filename: options };
     } else {
-      validateOptions(path.resolve(__dirname, '../schema/plugin.json'), options, 'Extract Text Plugin');
+      validateOptions(path.resolve(__dirname, './schema/plugin.json'), options, 'Extract Text Plugin');
```
evilebottnawi: Why?

src/index.js
```diff
@@ -40,6 +41,17 @@ class ExtractTextPlugin {
     return { loader: require.resolve('./loader'), options };
   }
 
+  static cloneModule(module) {
```
evilebottnawi: Can we move this in https://github.com/webpack-contrib/extract-text-webpack-plugin/blob/master/src/lib/helpers.js

src/index.js
```diff
@@ -145,25 +159,38 @@ class ExtractTextPlugin {
           const extractedChunk = extractedChunks[chunks.indexOf(chunk)];
           const shouldExtract = !!(options.allChunks || isInitialOrHasNoParents(chunk));
           chunk.sortModules();
-          async.forEach(chunk.mapModules(c => c), (module, callback) => { // eslint-disable-line no-shadow
+          async.forEach(chunk.mapModules((c) => { return c; }), (module, callback) => { // eslint-disable-line no-shadow, arrow-body-style
```
evilebottnawi: Why?

oviava:
does this have any more traction ? run some local tests with the forked branch and seems ok

src/index.js
BowlingX: ./schema is not resolved correctly when I build the branch from source. Not sure how the build process is, but without changing the path it does not compile.

src/index.js
BowlingX: can be changed to `c => c`, properly a missed merge

src/index.js
BowlingX: sure

FlorianH:
+1

src/index.js
danez: This should be reverted and the schema folder moved back to where it was.

src/index.js
```diff
@@ -88,7 +100,7 @@ class ExtractTextPlugin {
     if (Array.isArray(options) || isString(options) || typeof options.options === 'object' || typeof options.query === 'object') {
       options = { use: options };
     } else {
-      validateOptions(path.resolve(__dirname, '../schema/loader.json'), options, 'Extract Text Plugin (Loader)');
+      validateOptions(path.resolve(__dirname, './schema/loader.json'), options, 'Extract Text Plugin (Loader)');
```
danez: Should also be reverted.

BowlingX:
I will do the suggested changes and solve the conflicts today.

src/index.js
BowlingX: If I run the tests with ../schema I get:

    ENOENT: no such file or directory, open '/home/david/Projekte/extract-text-webpack-plugin/schema/loader.json'

src/index.js
```diff
@@ -145,28 +148,41 @@ class ExtractTextPlugin {
...
             let meta = module[NS];
             if (meta && (!meta.options.id || meta.options.id === id)) {
               const wasExtracted = Array.isArray(meta.content);
               // A stricter `shouldExtract !== wasExtracted` check to guard against cases where a previously extracted
               // module would be extracted twice. Happens when a module is a dependency of an initial and a non-initial
               // chunk. See issue #604
               if (shouldExtract && !wasExtracted) {
-                module[`${NS}/extract`] = shouldExtract; // eslint-disable-line no-path-concat
-                compilation.rebuildModule(module, (err) => {
+                const newModule = cloneModule(module);
```
michael-ciniawsky: `\n`

src/index.js
```diff
@@ -145,28 +148,41 @@ class ExtractTextPlugin {
...
+                newModule[`${NS}/extract`] = shouldExtract; // eslint-disable-line no-path-concat
```
michael-ciniawsky: `\n`

src/index.js
```diff
@@ -145,28 +148,41 @@ class ExtractTextPlugin {
...
+                compilation.buildModule(newModule, false, newModule, null, (err) => {
                   if (err) {
                     compilation.errors.push(err);
                     return callback();
                   }
-                  meta = module[NS];
+                  meta = newModule[NS];
```
michael-ciniawsky: `\n`

src/index.js
```diff
@@ -145,28 +148,41 @@ class ExtractTextPlugin {
...
+                  const identifier = module.identifier();
```
michael-ciniawsky: `\n`

src/index.js
```diff
@@ -145,28 +148,41 @@ class ExtractTextPlugin {
...
                   // Error out if content is not an array and is not null
                   if (!Array.isArray(meta.content) && meta.content != null) {
-                    err = new Error(`${module.identifier()} doesn't export content`);
+                    err = new Error(`${identifier} doesn't export content`);
                     compilation.errors.push(err);
                     return callback();
                   }
-                  if (meta.content) { extractCompilation.addResultToChunk(module.identifier(), meta.content, module, extractedChunk); }
+                  if (meta.content) {
```
michael-ciniawsky: `\n` (before)

src/index.js
```diff
@@ -194,6 +210,29 @@ class ExtractTextPlugin {
           callback();
         });
       });
+
+      compilation.plugin('optimize-module-ids', (modules) => {
+        modules.forEach((module) => {
+          const data = toRemoveModules[module.identifier()];
```
michael-ciniawsky: `\n`

src/index.js
```diff
@@ -194,6 +210,29 @@ class ExtractTextPlugin {
...
+          if (data) {
+            const oldModuleId = module.id;
+            const newModule = cloneModule(module);
```
michael-ciniawsky: `\n`

src/index.js
```diff
@@ -194,6 +210,29 @@ class ExtractTextPlugin {
...
+            newModule.id = oldModuleId;
+            newModule._source = data.module._source; // eslint-disable-line no-underscore-dangle
```
michael-ciniawsky: `\n`

src/index.js
```diff
@@ -194,6 +210,29 @@ class ExtractTextPlugin {
...
+            data.chunks.forEach((chunk) => {
+              chunk.removeModule(data.moduleToRemove);
```
michael-ciniawsky: `\n`

src/index.js
```diff
@@ -194,6 +210,29 @@ class ExtractTextPlugin {
...
+              const deps = data.moduleToRemove.dependencies;
```
michael-ciniawsky: `\n`

src/index.js
```diff
@@ -194,6 +210,29 @@ class ExtractTextPlugin {
...
+              deps.forEach((d) => {
+                if (d.module && d.module.loaders.length > 0) {
+                  chunk.removeModule(d.module);
+                }
+              });
```
michael-ciniawsky: `\n`

src/index.js
danez: Because you moved the schema into the src folder, but @michael-ciniawsky wants to have them in a different location anyway, see comment.

BowlingX:
I made the requested adjustments, Is there anything more needed here?

evilebottnawi:
@BowlingX Looks like we can close this PR, don't use `ETWP` for extract `css`, for this please use https://github.com/webpack-contrib/mini-css-extract-plugin. `ETWP` only for extract assets.

BowlingX:
@evilebottnawi I guess yes. I tested the new plugin and it works well for me. Also with the last changes (#604) my approach does not work anymore and I would have to refactor a bit. But properly it's unecessary because of the new plugin.

PR #557, opened 2017-06-29
Update package.json

pierreneter:


evilebottnawi:
@pierreneter already done, but only in beta https://github.com/webpack-contrib/extract-text-webpack-plugin/blob/v3.0.0-beta.3/package.json, just wait after https://github.com/webpack-contrib/extract-text-webpack-plugin/pull/540 was merged.

PR #553, opened 2017-06-24
refactor: Update deprecated `chunk.modules` functions

d3viant0ne:
Replaces remaining usage of the deprecated `chunk.modules` functions

- [x] refactor: Replace usage of chunk.modules

  BREAKING CHANGE: Updates to `Chunk.mapModules | forEachModule | getNumberOfModules`. This release is not backwards compatible with `Webpack 2.x` due to breaking changes in webpack/webpack#4764

Closes #548

PR #560, opened 2017-07-05
Add webpack 3.0 to acceptable peerDependencies

ajfuller:
I'm currently getting peer dependency warnings about this package when trying to use Webpack 3.0.
```
warning "extract-text-webpack-plugin@2.1.2" has incorrect peer dependency "webpack@^2.2.0".
```

Everything appears to be working as expected, so it should be safe to add 3.0 to the acceptable peerDependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "OrderUndefinedError.js"
   ],
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


michael-ciniawsky:
WIP in #540 && #553 Thx :)

PR #601, opened 2017-08-14
refactor: refer to the entry point instead of the first `module` (`module.identifier`)

akihikodaki:
The order of modules can be reordered by `optimize-module-order` plugins. The intention is to use the entry point as the default identifier, so refer to entries property of the compilation.


akihikodaki:
Commit 426e758b931fb67d711e25bcf57a5a5ad1c88102 adds a test case.

akihikodaki:
See also https://github.com/webpack/webpack.js.org/pull/1527.

akihikodaki:
https://github.com/miraiken/web/commit/46dfdde7694f232115b1d57bea7c0505cca13194 is an easy workaround.

evilebottnawi:
@akihikodaki thanks for PR

evilebottnawi:
@sokra need review

src/loader.js
```diff
@@ -94,13 +94,16 @@ export function pitch(request) {
       }
       try {
         let text = this.exec(source, request);
-        if (typeof text === 'string') { text = [[0, text]]; }
-        text.forEach((item) => {
-          const id = item[0];
-          compilation.modules.forEach((module) => {
-            if (module.id === id) { item[0] = module.identifier(); }
+        if (typeof text === 'string') {
+          text = [[compilation.entries[0].identifier(), text]];
```
sokra: You can use `[[entries[0].identifier(), text]]` too i guess.

evilebottnawi:
@akihikodaki friendly ping

michael-ciniawsky:
### `Issues`

- Fixes #648

sgal:
Also fixes https://github.com/webpack/webpack/issues/5733

akihikodaki:
52a194284a5dfd4c8a1543c82c0cd773ee447978 applies https://github.com/webpack-contrib/extract-text-webpack-plugin/pull/601#pullrequestreview-70527320.

d3viant0ne:
@akihikodaki - Once #652 lands, this will need one more quick rebase before it merges.

michael-ciniawsky:
@akihikodaki #652 just landed please rebase when time :)

akihikodaki:
8228819873bba36aded831f5b72e06189ac8c35a: rebased to 5286ab20ad83f6ade30192f939e0b42d3956d6f5.

michael-ciniawsky:
> Module build failed: TypeError: entries[0].identifier is not a function

@akihikodaki Can you revert to `compilation.entries[0].identifier` ?  Or create a new variable etc ?

akihikodaki:
f2c3056c25e35e807b74fa208ca50899c0af6708: Reverted the change. I reviewed the intension why I used `compilation.entries`, and it was actually because `entries` argument is not an array of the first modules to be executed, but of chunks.

michael-ciniawsky:
https://github.com/webpack-contrib/extract-text-webpack-plugin/blob/f2c3056c25e35e807b74fa208ca50899c0af6708/src/loader.js#L81 did you add this 🙃 , it seems to be unused ?

akihikodaki:
No, I didn't.

cletusw:
Thanks! Could someone post here when we get a release bump with this fix?

sgal:
Is there any plan to release a new version with this fix? The issue with `NamedModulesPlugin` and `HashedModuleIdsPlugin` is preventing users from implementing [long term caching solution](https://webpack.js.org/guides/caching/), so it would be nice to get it out.

Issue #604, opened 2017-08-16
Error: doesn't export content error every even rebuild


kostasmanionis:
I'm experiencing an issue where every even build throws this error in the console. By `every even build` I mean `Initial build -> Error thrown -> Rebuild Sucessful -> Error thrown -> Rebuild Sucessful`. The error is only visible in the browser console.

```
Uncaught TypeError: Cannot read property 'call' of undefined
    at __webpack_require__ (entryA.js:20)
    at Object.<anonymous> (entryA.js:9526)
    at __webpack_require__ (entryA.js:20)
    at Object.<anonymous> (entryA.js:9564)
    at __webpack_require__ (entryA.js:20)
    at Object.defineProperty.value (entryA.js:9555)
    at __webpack_require__ (entryA.js:20)
    at module.exports.ctor.super_ (entryA.js:63)
    at entryA.js:66
```

It basically tries to require a module with an `id` that doesn't exist. 

Here's an older issue that also describes the problem https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/42. Seems that @sokra correctly described what's going on https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/42#issuecomment-70731042. There's various fixes proposed there, though AFAIK this is the only one that really works is https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/42#issuecomment-271058702. 

I've set up a repo with a reproducible use case https://github.com/kostasmanionis/extract-text-plugin-rebuild-bug

Versions used:
```
    "css-loader": "^0.28.4",
    "extract-text-webpack-plugin": "^3.0.0",
    "style-loader": "^0.18.2",
    "webpack": "^3.5.4",
    "webpack-dev-server": "^2.7.1"
```

Here's a very minimal dependency graph that is able to reproduce this issue.

![screen shot 2017-08-16 at 15 32 28](https://user-images.githubusercontent.com/5301424/29363306-261c5960-8298-11e7-961d-b4c721fe7441.png)

PR #605, opened 2017-08-16
fix(index): stricter check for `shouldExtract !== wasExtracted`

kostasmanionis:
Tried this on a few projects that been having this issue and it works

### Issues
---
Fixes https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/604

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -149,7 +149,10 @@ class ExtractTextPlugin {
             let meta = module[NS];
             if (meta && (!meta.options.id || meta.options.id === id)) {
               const wasExtracted = Array.isArray(meta.content);
-              if (shouldExtract !== wasExtracted) {
+              // A stricter `shouldExtract !== wasExtracted` check to guard against cases where a previously extracted
+              // module would be extracted twice. Happens when a module is a dependency of an initial and a non-initial
+              // chunk. See issue #604
+              if (shouldExtract && !wasExtracted) {
                 module[`${NS}/extract`] = shouldExtract; // eslint-disable-line no-path-concat
                 compilation.rebuildModule(module, (err) => {
                   if (err) {


kostasmanionis:
I would love to add some tests, though I'm not really sure how to approach it 🤔  I would need to setup a webpack compiler, then trigger a rebuild and somehow check if the ids are correct... Or is there a simpler way?

evilebottnawi:
@kostasmanionis we have example of tests in `https://github.com/webpack-contrib/extract-text-webpack-plugin/tree/master/test`, without tests this PR most likely will not be accepted.

kostasmanionis:
@michael-ciniawsky there's some info and a reproducible use case in https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/604

I'm not yet sure how to test the rebuild functionality 🤔 If anyone has any suggestions, let me know

michael-ciniawsky:
@kostasmanionis Yep, sry I missed it :). hmm... current tests are passing and the change is minimal, async CSS chunks are a known problem with `ExtractTextPlugin`, I also don't know how to test for rebuilds tbh. Maybe add a comment above the check in case this f** up some other usecases and needs to be reverted :)

kostasmanionis:
@michael-ciniawsky tried to add a useful comment there, looks good?

evilebottnawi:
@kostasmanionis How difficult to implement the test?

kostasmanionis:
@evilebottnawi there's a reproducible repo https://github.com/kostasmanionis/extract-text-plugin-rebuild-bug

I'm thinking I would need to trigger a webpack rebuild to test this specific use case, since the problem does not occur on the initial build. Then I would need to check if all of the module ids are correct 🤔

evilebottnawi:
@kostasmanionis thanks, in near future i can investigate this - how we can implement test for this

evilebottnawi:
@kostasmanionis good work, your solution fix issue, but without tests we can change behavior (our just refactor code) in future and get regression :disappointed: Need search how we can implement tests.

kostasmanionis:
Thanks for finding time to look at it!

I have a few ideas on how to test it 🤔 Gonna try playing with a bit when I have some time, seems like a good opportunity to get a little bit more familiar with webpack internals.

Any idea when this fix could get published?

evilebottnawi:
@kostasmanionis in the near future :+1:

evilebottnawi:
@d3viant0ne need third approved :+1:

Issue #634, opened 2017-09-29
Generate one (or more) resulting files merging the original results


danikaze:
Proposal for a feature to solve the frequently asked question about _how to generate only one file with extract-text-webpack-plugin_.
Usually solved by external plugins (i.e. `merge-files-webpack-plugin`) in a completely different step, this doesn't look to be working with v3. Also, it would be nice to have a _native_ solution.

My proposal is to have an extra `merge` option to work like this, generalising the stated point:

```js
// Generate one styles.css from all the *.styles.css
// and one themes.css from all the *.themes.css files
new ExtractTextPlugin({
  merge: [{
    filename: 'styles.css',       // name of the merged output file
    test: /\.style\.css$/,        // files to include (*.styles.css)
    preventOriginalOutput: true,  // when true (by default), original files won't be generated
  }, {
    filename: 'themes.css',
    test: /\.theme\.css$/,
    preventOriginalOutput: true,
  }],
});
```

So let's say we have this files as original output from this plugin:
- `a.style.css`
- `b.style.css`
- `c.style.css`
- `a.theme.css`
- `b.theme.css`
- `fooBar.css`

The resulting files after the execution should be this:
- `styles.css` merging `a.style.css` + `b.style.css` + `c.style.css`
- `themes.css` merging `a.theme.css` + `b.theme.css`
- `fooBar.css` which doesn't merge into any other file

If we set `preventOriginalOutput` option to `false`, then the resulting file will output also the original files:
- `styles.css` merged from `a.style.css` + `b.style.css` + `c.style.css`
- `themes.css` merged from `a.theme.css` + `b.theme.css`
- `a.style.css`
- `b.style.css`
- `c.style.css`
- `a.theme.css`
- `b.theme.css`
- `fooBar.css`

d3viant0ne:
See comment in https://github.com/webpack-contrib/extract-text-webpack-plugin/pull/636#issuecomment-333866537

PR #636, opened 2017-09-29
feat: add option to merge files (`options.merge`)

danikaze:
Implements #634 

Current status: it's merging files properly but the map file is lost in the intermediate step.

    source1 --> file1.css + map1
    source2 --> file2.css + map2 --> mergedFile (no map)
    source3 --> file3.css + map3


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "extract-text-webpack-plugin",
-  "version": "3.0.0",
+  "version": "3.1.0",
```
michael-ciniawsky: Please revert this as we have release management in place and will take care of that

src/index.js
```diff
@@ -215,10 +235,41 @@ class ExtractTextPlugin {
...
-            compilation.assets[file] = source;
-            chunk.files.push(file);
+            let preventOutput = false;
+            this.options.merge.forEach((mergeChunk) => {
+              if (mergeChunk.test.test(file)) {
+                if (!this.filesToMerge[mergeChunk.filename]) {
+                  this.filesToMerge[mergeChunk.filename] = [];
+                }
+                this.filesToMerge[mergeChunk.filename].push(extractedChunk);
+                preventOutput = preventOutput || (mergeChunk.preventOriginalOutput !== false);
```
michael-ciniawsky: Can we find a better name (less verbose) then `preventOriginalOutput` => `prevent || original || preventOriginal || ?` ?

package.json
danikaze: understood

src/index.js
danikaze: sure, I was thinking also about that :P

I will include the tests, but wanted to make sure it wasn't going to become a dropped feature :)

michael-ciniawsky: 😛  Well we definitely need to check for ordering issues && multi entry setups and it needs approval from other maintainers aswell. But the feature itself is a nice improvement imho

danikaze:
Hello. It looks like test is passing now.
But it took me some time so I want to comment about the environment:
- `jest` is buggy. It doesn't allow any console.log to be printed. It's a known bug. Any workaround?
- Tests were running and passing in local environment, but not in appveyor.
- It's hard to debug because the running setup doesn't allow for breakpoints (or not easily) + no console.log = impossible to follow while running.
So, after just checking the logic, I made it work.
But travis is failing because of this

> ERROR: npm is known not to run on Node.js v4.3.2
> Node.js 4 is supported but the specific version you're running has
> a bug known to break npm. Please update to at least ${rel.min} to use this
> version of npm. You can find the latest release of Node.js at https://nodejs.org/

Anything to do?

michael-ciniawsky:
> jest is buggy. It doesn't allow any console.log to be printed. It's a known bug. Any workaround?

Could you post the e.g tracking issue or elaborate where/what isn't going to work? I use `console.log`with Jest on a regular basis and never have had any issue, but the ETWP code base might be different here :)

Does `npm t -- --verbose` work at least?

> Tests were running and passing in local environment, but not in appveyor

I will take a look later

> But travis is failing because of this

Ignore this for now as it is unrelated to ETWP or this PR in particular (it's a known `npm` issue with `node =< v4.5.0`)

danikaze:
`jest` bug: https://github.com/facebook/jest/issues/2166
It's not important, but when things doesn't work, it's hard to debug.

`appveyor` is passing now (after the last commit) so it should be ready to merge if we ignore the npm 4.3 issue in `travis`.
The main problem is, I was having a false positive in my local environment, don't know why because breakpoints weren't working.

d3viant0ne:
I want @sokra's take on this before it goes anywhere. Given this ( etwp ) is going away & being rolled back into Webpack, any feature added is one that has to be ported into Webpack.

My initial impression is conceptually, this is a bit murky, hard to sufficiently test & I guarantee people are going to have trouble with it. That said, if Tobias is okay with it conceptually then we can work through the implementation .

d3viant0ne:
@danikaze - We are going to have to pass on this one. Even if there wasn't already a plugin published for this, the last thing etwp needs at this point is more complexity.

This is already planned for retirement & the functionality rolled back into Webpack where it belongs, this particular use case may or may not be addressed at that time, it's not for me to say.

danikaze:
I see. Bad luck then.
I thought after finding lot of people looking for this feature and an outdated plugin, that would be nice to have this in the core of the plugin, but I see what you mean.
I´ll try to move the logic then to _yet another_ plugin, though it would be harder to generate map files in a complete different step, I think.

d3viant0ne:
@danikaze - If you feel there is enough demand for this particular plugin, we can discuss rolling it as a stand alone in contrib.

Given the state of etwp, where it's going & the similarities with the `allChunks` setting this isn't the right place for it.

danikaze:
@d3viant0ne sure.
Thanks for your time taking a look anyways ;)

PR #652, opened 2017-10-20
build(defaults): update to v1.6.0

d3viant0ne:
- Updates to minimum Node version of 4.8 ( npm 5 compatibility )
- Update devDependencies via `npm run webpack-defaults`
- Semantic changes to both CI configurations

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -5,7 +5,7 @@
       {
         "useBuiltIns": true,
         "targets": {
-          "node": "4.3"
+          "node": "4.8"
         },
         "exclude": [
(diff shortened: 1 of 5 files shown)


felixjung:
Why was this not listed as a breaking change in the release notes? Our build is now failing because of the Node version requirement 😢

michael-ciniawsky:
The node version bump itself is within the semver minor range, I know it's annoying in your case, but we needed to bump it bc of `npm@5` and we simply can't release any nitty-gritty as semver major. It's assumed that folks use the `@latest` within their current semver major range

PR #656, opened 2017-10-29
refactor(loader): export `locals` as ES2015 module

### LINKED ISSUE #229: Remove empty module functions in bundle
When using this plugin with `allChunks: true`, I get a number of empty module functions in the bundle:

```
function(module, exports) {

    // removed by extract-text-webpack-plugin

/***/ },
```

Is there a way to have webpack suppress/remove them to reduce the bloat?

glixlur:
### `Noteable Changes`

This pull request replaces `module.exports` with `export default` to allow `webpack` to take advantage of ES2015 module to reduce creating a scope for each stylesheet imported. It also solves #229 when ModuleConcatenationPlugin is turned on.

### `Needs Triage`

1. If you use CSS Module with a wildcard `import * as style from './style.css'` (which I don’t think was supported in the first place), it will stop working until `* as` is removed.
2. ~~Will this break for CommonJS users?~~ Yes.

### `Issues`

- Fixes #229

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -107,7 +107,7 @@ export function pitch(request) {
         }
         this[NS](text, query);
         if (text.locals && typeof resultSource !== 'undefined') {
-          resultSource += `\nmodule.exports = ${JSON.stringify(text.locals)};`;
+          resultSource += `\nexport default ${JSON.stringify(text.locals)};`;
         }
       } catch (e) {
         return callback(e);


src/loader.js
```diff
@@ -107,7 +107,7 @@ export function pitch(request) {
         }
         this[NS](text, query);
         if (text.locals && typeof resultSource !== 'undefined') {
-          resultSource += `\nmodule.exports = ${JSON.stringify(text.locals)};`;
+          resultSource += `\nexport default ${JSON.stringify(text.locals)};`;
```
michael-ciniawsky:
```js
resultSource += `\nexport const locals = ${JSON.stringify(text.locals)};`;
```

michael-ciniawsky: Needs triage which export is the best solution here

alexlur: What is the advantage of this over `export default`?

alexlur: Using default export means it works with all current CSS module imports (and hopefully avoid a major version bump?). With named export, we would have to change

```js
import app from './app.css'
import specific from './specific.sass'
```
to
```js
import { local as app } from './app.css'
import { local as specific } from './specific.sass'
```

michael-ciniawsky: I agree with you, it's not that I'm in favor of a named export, but I need to revisit the `css-loader` first to check

>  (and hopefully avoid a major version bump?)

Unfortunately unlikely because of  `require('./file.css)` (CJS) usage 😥 

```
require('./file.css') => require('./file.css').default
```

d3viant0ne: iirc, this is going to suffer from the same kind of issue that file-loader had.

Until the css-loader/new-loader this is only going to break things.

d3viant0ne: @alexlur - etwp will eventually be retired & it's functionality will be rolled back into webpack where it belongs.

This is a change that has to happen across the entire style sheet chain simultaneously and is slated to happen when it's all reworked to pull some of the missing style functionality back into webpack.

PR #660, opened 2017-11-04
fix: "Cannot read property 'initial' of undefined" (#202) (`webpack =< 1.0.0`)

revolunet:
copied from https://github.com/webpack-contrib/extract-text-webpack-plugin/pull/202/files

this fix the issue for webpack 1 projects


revolunet:
thanks. even on the webpack-1 branch ?

Issue #667, opened 2017-11-21
schema-utils validation errors not thrown properly


dinkzilla:
When validation in schema-util fails, schema-utils throws an error with fairly verbose messaging, but it is not being properly caught and rethrown within extract-text-webpack-plugin. As a result, you get unhelpful error messages in the console that look like this:

```
<project-path>\node_modules\schema-utils\dist\validateOptions.js:40
    throw new _ValidationError2.default(ajv.errors, name);
    ^

false
```

dinkzilla:
I created a pull request fixing the issue...

https://github.com/webpack-contrib/extract-text-webpack-plugin/pull/668

michael-ciniawsky:
Could you install the latest `schema-utils` manually and see if it works ? I fixed it in `schema-utils` directly with a nearly similar approach a few days ago

michael-ciniawsky:
Could you show a screenshot of the Error you get when using your patch instead please ? Adding the stack back to the `{Error}` was requested already :), I suppressed it in `schema-utils` tbh considering it noise when the Plugin Name etc is provided but in more complex setups it's unfortunately needed

dinkzilla:
It looks like this...

```
<poject-path>\node_modules\extract-text-webpack-plugin\index.js:131
                        throw new Error(e.message);
         ^
Error: Validation Error

Extract Text Plugin Invalid Options

options['sourceMap'] should NOT have additional properties

    at new ExtractTextPlugin (<project-path>\node_modules\extract-text-webpack-plugin\index.js:131:10)
    at Object.css (C:\Code\dearborn-station\DynamicPricing\node_modules\@easy-webpack\config-css\src\index.ts:33:25)
    at generateConfig (<project-path>\node_modules\@easy-webpack\core\src\index.ts:71:40)
<...full stack trace continues...>
```

I agree, it's potentially a lot of noise. Originally I wasn't including it either and was just doing a console.warn(e.message), but that didn't actually help me solve my particular issues. It's mainly an issue for people using easy-webpack I imagine, as we are in this project.

michael-ciniawsky:
We can limit it, the main issue is 

```
<project-path>\node_modules\extract-text-webpack-plugin\index.js:131
                        throw new Error(e.message);
```

[...]

Issue #17, opened 2014-09-15
Make extract-text-webpack-plugin re-entrant


andreypopp:
I'm working on React Style (see [experimental branch](https://github.com/andreypopp/webpack-stylegen) and also #8 is slightly related as I iterated form your suggestion).

TL;DR: it produces CSS from JS:

```
var ReactStyle = require('react-style')

var style = ReactStyle({
  color: 'red',
  backgroundColor: 'white'
})
```

Becomes:

```
.automatically_generated_class_name {
  color: 'red';
  background-color: 'white';
}
```

.
The problem arises when I use both `css-loader` and `react-style-loader` with `extract-text-webpack-plugin` (I need to combine both approaches b/c some legacy styles I still want to import via CSS modules, like bootstrap for example).

Because plugin compiles mini-bundle which includes all JS code (due to `react-style-loader`) it encounters css requires once more but b/c child compiler has no `extract-text-webpack-plugin` configured for itself but inherits all loaders from main compiler it fails. But more importantly I don't want to process CSS files during react style bundle compilation b/c such CSS files will be processed by another child compiler (a direct child of main compiler).

So my question is how to make/configure `extract-text-webpack-plugin` re-entrant so child compilers don't produce another child compilers.

PR #668, opened 2017-11-21
fix(package): update `schema-utils` v0.3.0...0.4.1

dinkzilla:
### `Issues`

- Fixes for https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/667

### `Notable Changes`

This pull request properly catches and throws validation errors within schema-utils so that they appear appropriately in the console.

As an example, a validation error which used to read as follows

```
<project-path>\node_modules\schema-utils\dist\validateOptions.js:40
    throw new _ValidationError2.default(ajv.errors, name);
    ^

false
```

now reads e.g

```
Extract Text Plugin Invalid Options

options['sourceMap'] should NOT have additional properties
```

### DIFF
--- a/test/schema-validation.test.js
+++ b/test/schema-validation.test.js
@@ -15,9 +15,13 @@ describe('json schema validation', () => {
     }).doesNotThrow;
   });
 
-  it('throws if an incorrect config is passed in', () => {
+  it('displays error if an incorrect config is passed in', () => {
     const exit = jest.spyOn(process, 'exit').mockImplementation(() => { });
+    const consoleError = jest.spyOn(console, 'error').mockImplementation(() => { });
+
     ExtractTextPlugin.extract({ style: 'file.css' });
+
+    expect(consoleError).toHaveBeenCalled();
     expect(exit).toHaveBeenCalledWith(1);
   });
 });
--- a/test/schema-validation.test.js
+++ b/test/schema-validation.test.js
@@ -16,8 +16,8 @@ describe('json schema validation', () => {
   });
 
   it('throws if an incorrect config is passed in', () => {
-    expect(() => {
-      ExtractTextPlugin.extract({ style: 'file.css' });
-    }).toThrow();
+    const exit = jest.spyOn(process, 'exit').mockImplementation(() => { });
+    ExtractTextPlugin.extract({ style: 'file.css' });
+    expect(exit).toHaveBeenCalledWith(1);
   });
 });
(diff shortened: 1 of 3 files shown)


dinkzilla:
Based on conversation in the #667 thread with @michael-ciniawsky, The correct fix was actually to update to the newest version of schema-utils, which now logs errors to the console. I have updated this pull request accordingly. Schema-utils will now log the error to the console and then exit, which required updating the validation failure test.

test/schema-validation.test.js
```diff
@@ -15,9 +15,13 @@ describe('json schema validation', () => {
     }).doesNotThrow;
   });
 
-  it('throws if an incorrect config is passed in', () => {
```
michael-ciniawsky: I'm sry about this 😅 I should have taken a look sooner but could you add `process.env.JEST = true` on L3-4 instead so `schema-utils` will rethrow the `{Error}` instead of exiting while running tests ? The 'old' test should work then. 

```js
const err = () => new ExtractTextPlugin({ filename: 1 })

expect(err).toThrow()
// Test if we get the correct err.message
expect(err).toThrowErrorMatchingSnapshot()
```

dinkzilla: But in reality, doesn't it write to console and exit, not throw? My test was written to match that scenario. 

The test you describe would essentially be "when in testing with a bad config, does it throw?" Unless I'm missing something, that doesn't seem like a very valuable test.

(Not sure why one of the checks is failing, I've been on holiday and but tests were passing on my computer before I left. Can look into this tomorrow, and will update based on your response.)

michael-ciniawsky: The test should be changed back, but please hold on for ~1-2 days we resolved this in [webpack core (#6035)](https://github.com/webpack/webpack/pull/6035) and [`schema-utils` (#17)](https://github.com/webpack-contrib/schema-utils/pull/17)

evilebottnawi:
@dinkzilla friendly ping :+1:

dinkzilla:
You caught me, been busy and totally forgot this was out here. I'll try to get to it this weekend. Thanks for the ping @evilebottnawi

test/schema-validation.test.js
dinkzilla: @michael-ciniawsky Can you explain why the test should be changed back? I am still confused about what it is testing.

michael-ciniawsky: It's tests for malformed options based on a JSON Schema and should work without the need to mock anything in the test, please revert the test changes and I will finally land this on `master` :)

michael-ciniawsky:
I will update it myself with the next patch release

PR #707, opened 2018-01-28
refactor: update to new plugin system and API's

horo42:
The plugin stopped working with webpack 4 because of breaking changes. I hacked away at it for a couple of hours and managed to get it to build without errors.

I haven't seen a lot of activity here, so I figured I'd share my work and maybe spare someone some time.


TheLarkInn:
Wow! Thank you!!!

src/lib/helpers.js
```diff
@@ -1,5 +1,10 @@
 export function isInitialOrHasNoParents(chunk) {
-  return chunk.isInitial() || chunk.getParents().length === 0;
+  const parentCount = Array.from(
+    chunk.groupsIterable,
+    (chunkGroup) => chunkGroup.getParents().length
+  ).reduce((current, sum) => current + sum);
```
sciyoshi: This was giving me an error:
```
TypeError: Reduce of empty array with no initial value
    at Array.reduce (<anonymous>:null:null)
    at isInitialOrHasNoParents (/home/sciyoshi/workspace/web/webpack/extract-text-webpack-plugin/dist/lib/helpers.js:15:102)
```
Changing the line to `).reduce((current, sum) => current + sum, 0);` fixed it.

horo42: Thank you! I noticed that just yesterday, along with failing integration test. I didn’t push yet because I couldn’t make it pass everything.

I’ll fix and push the obvious low hanging fruits I missed and will keep working on the tests.

Expect an update ~12 hours later.

evilebottnawi: @horo42 also can you add tests around this?

src/index.js
```diff
@@ -168,19 +201,20 @@ class ExtractTextPlugin {
                 extractedChunk.addGroup(chunkGroup);
               }
```
horo42: We add the ChunkGroups to the extracted chunk and the _children and _parents sets belong to ChunkGroup now. Please confirm that these lines are unnecessary.

horo42: We add the ChunkGroups to the extracted chunk and the _children and _parents sets belong to ChunkGroup now. Please confirm that these lines are unnecessary.

src/index.js
```diff
@@ -122,83 +152,119 @@ class ExtractTextPlugin {
...
+            const extractedChunk = extractedChunks[i];
+            extractedChunk.index = i;
+            extractedChunk.originalChunk = chunk;
+            extractedChunk.name = chunk.name;
+
+            for (const chunkGroup of chunk.groupsIterable) {
+              if (chunkGroup instanceof Entrypoint) {
+                extractedChunk.addGroup(chunkGroup);
+              }
+
```
horo42: We add the ChunkGroups to the extracted chunk and the _children and _parents sets belong to ChunkGroup now. Please confirm that these lines are unnecessary.

horo42:
Test output: https://pastebin.com/z5FiXqk7

I'm not entirely sure where to look for these bugs. Guidance is more than welcome.
Will keep looking in my free time.

src/index.js
michael-ciniawsky: I can't 💯 atm 😛, but likely that this can be removed with the introduction of `ChunkGroups` now

michael-ciniawsky:
@horo42 I will chime in deeper into this PR in a few days, feel free to leave comments/questions as so see fit i will try to work through them then :)

TheLarkInn:
@michael-ciniawsky @sokra @ooflorent would one of you give this a quick pass rereview? I'd like to merge this to next branch and have a -next build shipped for this if theres no blockers on it.

horo42:
According to the integration tests this PR breaks async imports (sometimes?) and doesn't respect css import order. I'm pretty sure the issue lies in the merging of the chunks.

The ["common-async"](https://circleci.com/gh/webpack-contrib/extract-text-webpack-plugin/66) test (2nd failing test) fails because at the very end of the build process the plugin emits the same file twice with the contents "a\nb" and "a\nc", thus overwriting it.

That said, if I'm not using async imports it works fine, and there's precedent to this plugin being published while breaking css import order. (not that it's OK)

src/index.js
```diff
@@ -1,16 +1,17 @@
...
 import path from 'path';
 import Chunk from 'webpack/lib/Chunk';
+import Entrypoint from 'webpack/lib/Entrypoint';
 import { ConcatSource } from 'webpack-sources';
 import async from 'async';
 import loaderUtils from 'loader-utils';
 import validateOptions from 'schema-utils';
 import ExtractTextPluginCompilation from './lib/ExtractTextPluginCompilation';
-import OrderUndefinedError from './lib/OrderUndefinedError';
+// import OrderUndefinedError from './lib/OrderUndefinedError';
```
ooflorent: Could you remove the commented lines?

src/index.js
```diff
@@ -58,34 +59,59 @@ class ExtractTextPlugin {
...
-        if (isInitialOrHasNoParents(c)) return;
-        this.mergeNonInitialChunks(c, intoChunk, checkedChunks);
-      }, this);
+
+      // for (const chunkModule of chunk.modulesIterable) {
+      //   intoChunk.addModule(chunkModule);
+      //   chunkModule.addChunk(intoChunk);
+      // }
+
+      for (const chunkGroup of chunk.getChildren()) {
```
ooflorent: Use `childrenIterable`

src/index.js
```diff
@@ -58,34 +59,59 @@ class ExtractTextPlugin {
...
+        for (const chunkGroupChunk of chunkGroup.getChildren()) {
```
ooflorent: Same here

src/index.js
```diff
@@ -58,34 +59,59 @@ class ExtractTextPlugin {
...
+          for (const chunkModule of chunkGroupChunk.getModules()) {
```
ooflorent: Use `modulesIterable`

src/index.js
```diff
@@ -58,34 +59,59 @@ class ExtractTextPlugin {
...
+      // }, this);
     }
   }
 
   renderExtractedChunk(chunk) {
     const source = new ConcatSource();
-    chunk.forEachModule((module) => {
-      const moduleSource = module.source();
+
+    for (const chunkModule of chunk.getModules()) {
```
ooflorent: Use `modulesIterable`

src/index.js
```diff
@@ -122,83 +152,119 @@ class ExtractTextPlugin {
 
   apply(compiler) {
     const options = this.options;
+
     compiler.plugin('this-compilation', (compilation) => {
```
ooflorent: Can you use `hooks.thisCompilation` here please?

src/index.js
```diff
@@ -122,83 +152,119 @@ class ExtractTextPlugin {
...
-                content
-              )}`
-            );
-          }
-          module[NS] = {
-            content,
-            options: opt || {},
+
+      compilation.hooks.normalModuleLoader.tap(
+        'normal-module-loader',
```
ooflorent: The tap name should be the name of the caller. Here it's `ExtractTextPlugin`

src/index.js
```diff
@@ -122,83 +152,119 @@ class ExtractTextPlugin {
...
+        (loaderContext, module) => {
+          loaderContext[NS] = (content, opt) => {
+            if (options.disable) {
+              return false;
+            }
+
+            if (!Array.isArray(content) && content != null) {
+              throw new Error(
+                `Exported value was not extracted as an array: ${JSON.stringify(
+                  content
```
ooflorent: Formatting is weird.

src/index.js
```diff
@@ -122,83 +152,119 @@ class ExtractTextPlugin {
...
+          extractedChunks = chunks.map(() => new Chunk());
+
+          chunks.forEach((chunk, i) => {
+            const extractedChunk = extractedChunks[i];
+            extractedChunk.index = i;
+            extractedChunk.originalChunk = chunk;
+            extractedChunk.name = chunk.name;
+
+            for (const chunkGroup of chunk.groupsIterable) {
+              if (chunkGroup instanceof Entrypoint) {
```
ooflorent: Could you change this test to `chunkGroup.isInitial()`?

src/index.js
```diff
@@ -208,93 +274,83 @@ class ExtractTextPlugin {
...
-                        module.identifier(),
-                        meta.content,
-                        module,
-                        extractedChunk
-                      );
                     }
-                    callback();
+                  } else callback();
+                },
+                (err) => {
```
ooflorent: Can't you directly use `callback` here?

src/index.js
```diff
@@ -208,93 +274,83 @@ class ExtractTextPlugin {
...
-                );
+                }
+              });
+              compilation.hooks.optimizeExtractedChunks.call(extractedChunks);
+              callback();
+            }
+          );
+        }
+      );
+      compilation.hooks.additionalAssets.tapAsync(
```
ooflorent: You don't need to be `async` here. You can use `tap` and remove `callback` calls.

ooflorent:
Nice work! Could you have a look at the requested changes?
tl;dr:

- use `*Iterable` instead of `get*`. The getters drain the iterable so it is walked twice
- when you _tap_ into a hook, the name is _who is tapping_. Here it's `ExtractTextPlugin`.

horo42:
Thank you for your review. I will apply your recommendations to the PR tonight (in the next 4-6 hours).

Let me know if there’s anything else, and please keep an eye out for the update. The more eyes the better. :)

Cheers.

michael-ciniawsky:
Once the requested changes by @ooflorent are addressed I will cut an `alpha` release with this PR so folks can start trying it out and outstanding fixes can simply follow up

src/index.js
michael-ciniawsky:
```js
class ExtractTextPlugin {
   constructor() {
     this.plugin = { name: 'ExtractTextPlugin' };
     ...
   }

   apply(compiler) {
     const { plugin } = this;
     const { someHook } = compiler.hooks;
     
     someHook.tap(plugin, () => {});
   }
}
```

src/index.js
michael-ciniawsky: Please leave them for now I will tinker with this code once the PR landed :)

src/index.js
```diff
@@ -208,93 +274,83 @@ class ExtractTextPlugin {
...
+        'additional-assets',
+        (callback) => {
+          extractedChunks.forEach((extractedChunk) => {
+            if (extractedChunk.getNumberOfModules()) {
+              extractedChunk.sortModules((a, b) => {
```
michael-ciniawsky: `sortModules` is deprecated aswell, not to important though (leave it if it anything else causes failures) but if you can refcator it to use a standard `Array.prototype.sort` instead it would be great 👍. Maybe `chunk.getModules().sort()` works here

appveyor.yml
```diff
@@ -7,13 +7,13 @@ init:
 environment:
   matrix:
     - nodejs_version: '8'
-      webpack_version: latest
+      webpack_version: '4.0.0-beta.0'
```
michael-ciniawsky: Please don't alter this, we have a boilerplate package (`webpack-defaults`) which automatically updates these files

src/loader.js
```diff
@@ -8,16 +8,18 @@ import SingleEntryPlugin from 'webpack/lib/SingleEntryPlugin';
 import LimitChunkCountPlugin from 'webpack/lib/optimize/LimitChunkCountPlugin';
 
 const NS = path.dirname(fs.realpathSync(__filename));
+const pluginName = 'ExtractTextPlugin';
```
michael-ciniawsky: `const plugin = { name: 'ExtractTextPlugin' }`

appveyor.yml
michael-ciniawsky: What is not working here atm ?

horo42: Sorry about that 😅 I'll revert that change.

src/index.js
horo42: Reverting this change, because it broke the whole plugin 😨
This whole section is a jumbled up mess of asynchrony.
Nevertheless, I'm wholly behind the idea of refactoring, but first I'd like to see the core functionality fixed.

src/index.js
```diff
@@ -57,35 +56,51 @@ class ExtractTextPlugin {
...
-  renderExtractedChunk(chunk) {
+  static renderExtractedChunk(chunk) {
     const source = new ConcatSource();
-    chunk.forEachModule((module) => {
-      const moduleSource = module.source();
+
+    for (const chunkModule of chunk.modulesIterable) {
+      let moduleSource = chunkModule.source();
+
+      if (moduleSource instanceof CachedSource) {
```
horo42: During my testing I noticed that async imports are CachedSource instances caching a ReplaceSource instance, which breaks the plugin.

The issue probably lies elsewhere.

src/index.js
```diff
@@ -104,168 +119,191 @@ class ExtractTextPlugin {
...
-            extractedChunk.sortModules((a, b) => {
-              if (!options.ignoreOrder && isInvalidOrder(a, b)) {
-                compilation.errors.push(
-                  new OrderUndefinedError(a.getOriginalModule())
-                );
-                compilation.errors.push(
-                  new OrderUndefinedError(b.getOriginalModule())
-                );
-              }
-              return getOrder(a, b);
```
horo42: The getOrder helper function uses the underlying index2 property of modules. Unfortunately this property seems to be set by webpack after [sealing](https://webpack.js.org/api/compilation/#seal) and apparently sealing happens after the [additionalAssets](https://webpack.js.org/api/compilation/#additional-assets-async) hook, which is where the sorting takes place, because accessing index2 yielded null every time.

horo42:
🎉 Tests:       1 failed, 37 passed, 38 total

The only remaining failing test is the CSS import order test case. I guess it fails because the test assumes a depth first order and webpack does it breadth first.

src/loader.js
```diff
@@ -49,33 +52,39 @@ export function pitch(request) {
...
-          module.loaders = loaders.map((loader) => {
-            return {
-              loader: loader.path,
-              options: loader.options,
-            };
-          });
+    // a child compiler so we don't spawn other child compilers from there.
+    childCompiler.hooks.thisCompilation.tap(plugin.name, (compilation) => {
+      compilation.hooks.normalModuleLoader.tap(
+        plugin.name,
```
michael-ciniawsky: You can just pass the `{Object}` here, since in theory `plugin` could contain `{ name: 'MyPlugin', before: 'SomePlugin', after: 'SomePlugin' }`

src/index.js
michael-ciniawsky: Please add a comment for this to the code :)

src/index.js
michael-ciniawsky: Please add a comment for this to the code :)

src/index.js
horo42: I checked the implementation of ChunkGroup, isInitial() always returns false. There’s a deprecation warning that suggests using the instanceof Entrypoint check.

TheLarkInn:
@michael-ciniawsky Looks like changes have been added. Are these enough to ship alpha?

src/index.js
```diff
@@ -57,35 +59,60 @@ class ExtractTextPlugin {
...
-      source.add(
-        this.applyAdditionalInformation(
-          moduleSource,
-          module.additionalInformation
-        )
-      );
-    }, this);
+
+    for (const chunkModule of chunk.modulesIterable) {
+      let moduleSource = chunkModule.source();
```
log2-hwan: Seems this code causes runtime error when chunkModule is `ConcatenatedModule` instance (https://github.com/webpack/webpack/blob/master/lib/optimize/ConcatenatedModule.js#L370). `ConcatenatedModule`'s `source` function requires dependencyTemplate and runtimeTemplate parameters.

horo42: Can you please share the code that causes this runtime error when compiled? If the code isn’t public, could you please try removing as much code as possible so as to still exit with a runtime error?

Thanks!

michael-ciniawsky: @log2-hwan Please open an issue about this with some logs if possible

michael-ciniawsky:
https://github.com/webpack-contrib/extract-text-webpack-plugin/releases/tag/v4.0.0-alpha.0 🎉

Issue #642, opened 2017-10-12
get wrong css order after using HashedModuleIdsPlugin


silaike:
**Do you want to request a *feature* or report a *bug*?**
Bug.

**What is the current behavior?**
get wrong css order after using HashedModuleIdsPlugin

**If the current behavior is a bug, please provide the steps to reproduce.**
webpack plugin configs.
```javascript
       ...
	plugins: [
		new HtmlWebpackPlugin( {
			filename: 'index.html',
			template: 'index.ejs',
			minify: {
				collapseWhitespace: true,
				minifyJS: true
			}
		} ),

		new ExtractTextPlugin( {
			filename: 'css/[name].[contenthash:8].css',
			allChunks: true,
		} ),

		new OptimizeCssAssetsPlugin( {
			cssProcessor: require( 'cssnano' ),
			cssProcessorOptions: {
				autoprefixer: false
			}
		} ),

		new webpack.HashedModuleIdsPlugin(),

		new webpack.optimize.CommonsChunkPlugin( {
			names: [ 'vendor', 'manifest' ],
		} ),

		new webpack.optimize.CommonsChunkPlugin( {
			async: 'async-common',
			minChunks: ( module, count ) => (
				count >= 2
			),
		} ),

		new webpack.DefinePlugin( {
			'process.env': {
				NODE_ENV: JSON.stringify( 'production' )
			}
		} ),

		new webpack.optimize.UglifyJsPlugin( {
			compress: {
				warnings: false
			},
			comments: false
		} ),

		new webpack.optimize.ModuleConcatenationPlugin(),
	]
```

**What is the expected behavior?**
remove HashedModuleIdsPlugin then css order will be right.

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**
webpack version v3.6.0

silaike:
when i set `allChunks: false` and set `fallback: "style-loader"` , it works fine too, so I think its extract-text-webpack-plugin's issue.

evilebottnawi:
@silaike can you create minimum reproducible test repo?

symbios-zi:
the same problem at my styles. 
When I import in my app.scss dependesies, bootstrap is loading as last. It is wrong. The file called "AdminLTE.css" must loads as last
```
@import '../../../../public/static/plugins/bootstrap/assets/stylesheets/bootstrap.scss';

@import '../../../../public/static/plugins/AdminLTE/css/skins/skin-blue.css';
[...]

PR #709, opened 2018-02-01
test(cases): hashed module ids

evilebottnawi:
### `Issues`

- Closes https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/642


evilebottnawi:
@michael-ciniawsky I think we can merge this PR as is, because it is only tests :smile:

PR #712, opened 2018-02-01
fix(loader): change comment type

sntiagoma:
Double slash (//) comment causes problems with other css loaders.
Instance is used /**/ that's fully compatible


evilebottnawi:
@sntiagoma Can you provide url which loader have problematic?

PR #680, opened 2017-12-09
docs(README): HMR (`extracted-loader|css-hot-loader`)

sheerun:
Disclaimer: I'm an author of extracted-loader

PR #714, opened 2018-02-04
feat(loader): add support for hot reloading (`options.hmr`)

sheerun:
Closes #680

The hot reloading code live tested and the same as in https://github.com/sheerun/extracted-loader

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -17,15 +17,16 @@ if (module.hot) {
   var injectCss = function injectCss(prev, href) {
     var link = prev.cloneNode();
     link.href = href;
-    link.onload = function() {
+    link.onload = link.onerror = function() {
       prev.parentNode.removeChild(prev);
     };
     prev.stale = true;
-    prev.parentNode.insertBefore(link, prev);
+    prev.parentNode.insertBefore(link, prev.nextSibling);
   };
   module.hot.dispose(function() {
     window.__webpack_reload_css__ = true;
   });
+  module.hot.accept();
   if (window.__webpack_reload_css__) {
     module.hot.__webpack_reload_css__ = false;
     console.log("[HMR] Reloading stylesheets...");
@@ -62,7 +63,7 @@ export function pitch(request) {
     let resultSource;
     if (query.remove) {
       resultSource = '// removed by extract-text-webpack-plugin';
-      if (process.env.NODE_ENV === 'development') {
+      if (query.hot || query.hmr || this.hot) {
         resultSource += hmrCode();
       }
     } else {
--- a/src/loader.js
+++ b/src/loader.js
@@ -11,6 +11,36 @@ const NS = path.dirname(fs.realpathSync(__filename));
 
 export default source => source;
 
+function hmrCode() {
+  return (`
+if (module.hot) {
+  var injectCss = function injectCss(prev, href) {
+    var link = prev.cloneNode();
+    link.href = href;
+    link.onload = function() {
+      prev.parentNode.removeChild(prev);
+    };
+    prev.stale = true;
+    prev.parentNode.insertBefore(link, prev);
+  };
+  module.hot.dispose(function() {
+    window.__webpack_reload_css__ = true;
+  });
+  if (window.__webpack_reload_css__) {
+    module.hot.__webpack_reload_css__ = false;
+    console.log("[HMR] Reloading stylesheets...");
+    var prefix = document.location.protocol + '//' + document.location.host;
+    document
+      .querySelectorAll("link[href][rel=stylesheet]")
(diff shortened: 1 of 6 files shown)


src/loader.js
```diff
@@ -32,6 +62,9 @@ export function pitch(request) {
     let resultSource;
     if (query.remove) {
       resultSource = '// removed by extract-text-webpack-plugin';
+      if (process.env.NODE_ENV === 'development') {
```
evilebottnawi: Let's use `hot` option for this :+1:

sheerun:
I have no idea how to test it

evilebottnawi:
@sheerun Can you create simple example to easy test this?

src/loader.js
```diff
@@ -32,6 +62,9 @@ export function pitch(request) {
...
+        resultSource += hmrCode();
```
michael-ciniawsky: This would include the HMR Code for every extracted module right? I don't think this should be the case, but also didn't test it yet

src/loader.js
michael-ciniawsky: `options.hmr || options.hot` or `webpack >= v4.0.0` also adds `this.hot` to the `loaderContext` now

michael-ciniawsky: `webpack` loader/plugins shouldn't directly rely on `process.env` whenever possible

src/loader.js
```diff
@@ -11,6 +11,36 @@ const NS = path.dirname(fs.realpathSync(__filename));
...
+  });
+  if (window.__webpack_reload_css__) {
+    module.hot.__webpack_reload_css__ = false;
+    console.log("[HMR] Reloading stylesheets...");
+    var prefix = document.location.protocol + '//' + document.location.host;
+    document
+      .querySelectorAll("link[href][rel=stylesheet]")
+      .forEach(function(link) {
+        if (!link.href.match(prefix) || link.stale) return;
+        injectCss(link, link.href.split("?")[0] + "?unix=${+new Date()}");
```
michael-ciniawsky:
```js
"?unix=${+new Date()}"
```

What's the purpose for this query param ? Is it needed ?

src/loader.js
evilebottnawi: @sheerun don't rely on `NODE_ENV`, just add `hot` flag :+1:

sheerun: any idea how to output it just one time?

src/loader.js
sheerun: yes, this is needed so hmr recognizes extracted stylesheet as modified and performs reload

src/loader.js
sheerun: @michael-ciniawsky When I run `../node_modules/.bin/webpack-dev-server --watch --hot` both `query` and `this.options` are missing either "hot" or "hmr". the "this.hot" is also undefined.

Are you sure this is the way?

sheerun: Oh, I get it, you've decided to make it an option? @evilebottnawi told me before it should a default

sheerun:
@evilebottnawi @michael-ciniawsky Summary of changes:

1. The hot module reloading is (sadly) disabled by default (as requested), and can be enabled with query.hot, query.hmr of loader options, or this.hot that is supposed to work in webpack 4
2. Manual "test" is added: `npm run start:example` and modify style.css, it should hotreload

I hope it fulfills your requirements

evilebottnawi:
@sheerun Great works! Thanks!
@michael-ciniawsky what do you think about one options (example `hmr` as in `style-loader`)?

michael-ciniawsky:
@evilebottnawi Yep, it should be one, but not sure which is 'better', `style-loader` uses `hmr`, but the webpack API is called `hot` (which maybe a good idea to align with), `hmr` is more descriptive though imho

⚠️ This still needs some testing in general and since `ETWP` can extract other contents aswell, we need to think about adding this carefully, maybe it better suited for `css-webpack-plugin` (I will test it out) and the `next` branch also needs to be tested with this and the PR ported then

michael-ciniawsky:
This plugin has entered maintenance mode and won't receive any further feature updates as it is superseded by `mini-css-extract-plugin` until 'native' CSS supports lands in `webpack` directly

HMR is an important TODO in `mini-css-extract-plugin` atm if you have any ideas here feel free to send a PR there :)

If anyone has strong objections to still add HMR to `extract-text-webpack-plugin` nevertheless, we may reconsider this, but ETWP is the #1 on the list of things that should become obsolete asap so it maybe better to just let it RIP :)

sheerun:
@michael-ciniawsky It would make it easier to implement HMR for mini-css-extract-plugin if you added example project in there

PR #748, opened 2018-03-14
update this plugin to support webpack4

NealST:
Recently,I updated my webpack version to v4,but I encounter a problem that current extract-text-webpack-plugin does not support webpack 4, so I made a fork to solve this problem, after verifying my change support v4 successfully, I want to create this pull request to the main repository to share  my endeavor. Please contact me if any problem appear


evilebottnawi:
@NealST please use https://github.com/webpack-contrib/extract-text-webpack-plugin/releases/tag/v4.0.0-beta.0 for `webpack@4`

PR #769, opened 2018-04-03
recommend to use mini-css-extract-plugin for webpack 4.0+

epicfaace:
Otherwise, poor developers like me will only figure this out from googling and reading through the issues on this repository.


evilebottnawi:
Thanks for PR close in favor https://github.com/webpack-contrib/extract-text-webpack-plugin/pull/771

PR #813, opened 2018-12-03
[Minor/ Optimization] - forEach doesn't actually break/ return

MichaelWalker-git:
**Feature**
Changed the forEach to a for..of loop to allow it to exit early.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

**Feedback**
I am open to suggestions/ improvements, since I did delete the explicit this context.


alexander-akait:
DEPRECATED
Please use: https://github.com/webpack-contrib/mini-css-extract-plugin

